PR #6, opened 2015-02-25
added initialisation of intbv with 'plain text' string

josyb:
Code to initialise an intbv with a 'plain text' string, e.g.:
```python
intbv('MyHDL0.9')
```
will generate this:
VHDL:
```VHDL
control_rdr_self_dr(512-1 downto 448) <= unsigned'( "0011100100101110001100000100110001000100010010000111100101001101" );
```
Verilog:
```Verilog
assign  control_rdr_self_dr[512-1:448] = 4120283813211830605;
````


jandecaluwe:
Please explain why this would be useful, I don't see it.

josyb:
E.g. when implementing a USB function you have things like string descriptors which are essentially strings.
I also tend to have some text inside streams of data, especially in protocols, that help one decode 'on the fly'. 
Of course the same effect can be obtained via a local user function. The proposed 'addition' doesn't break anything, though.

jandecaluwe:
Mm, if there is a single issue causing maintenance headaches in with Py3K it is exactly the semantic changes regarding strings. As you say, this can easily be done locally and I feel that is a cleaner and clearer solution.

josyb:
OK, I'll move the change into my local 'utilities'.
And close this PR

PR #17, opened 2015-03-02
added py.test test/core2 to travis

cfelton:


jandecaluwe:
About test_AssignSignal - this tests an experimental assign feature that I never really liked. Is it OK to remove this already from the test suite?

cfelton:
I am ok with the test being removed, I assume we are keeping the py.test (core2) type tests.

jandecaluwe:
Yes, they are based on py.test and therefore easier. I will be extending them soon.

Issue #12, opened 2015-02-27
Problem simulating process with intbv signals of width 1


josyb:
In my registerfile module I have these processes:
```python
import os
from myhdl import *
import hdlutils, SimulateAvalon

def regrw(OFFSET, LENGTH, START, WIDTH, Clk, Reset, A, WD, Wr, Q, Pulse = None):
    ''' the WriteRead register '''

    @always_seq( Clk.posedge, reset = Reset)
    def ccregrw():
        if Wr and (A == OFFSET):
            Q.next = WD[WIDTH+START:START]

    return ccregrw

def ccassign( D, Q):
    ''' to assign the output ports from internal signals '''
    @always_comb
    def cca():
        Q.next = D

    return cca

def sim_intbv_width_1(Clk, Reset, A, WD, Wr, Rd, RQ, TestBit, TestVector):
    """   """
    lq1 = Signal(intbv(0)[1:])
    lq2 = Signal(intbv(0)[2:])
    reg1 = regrw(0,1,0,1, Clk, Reset, A, WD, Wr, lq1)
    reg2 = regrw(1,1,0,2, Clk, Reset, A, WD, Wr, lq2)
    ao1 = ccassign( lq1 , TestBit )
    ao2 = ccassign( lq2 , TestVector )

    rbd = ConcatSignal( lq2, lq1)

    @always_seq(Clk.posedge, reset = Reset)
    def mmrdr():
        if Rd:
            RQ.next = 0
            if A == 0 :
                RQ.next = rbd[1:]
            elif A == 1:
                RQ.next = rbd[:1]

    return instances()


def test_sim_intbv_width_1():

    hw_inst = sim_intbv_width_1(Clk, Reset, A, WD, Wr, Rd, RQ, TestBit, TestVector)

    ClkCount = Signal( intbv( 0 )[32:])
    tCK = 20

    @instance
    def clkgen():
        yield hdlutils.genClk(Clk, tCK, ClkCount)


    @instance
    def resetgen():
        yield hdlutils.genReset(Clk, tCK, Reset)

    @instance
    def stimulus():
        yield hdlutils.delayclks(Clk, tCK, 10)
        for i in range(2):
            yield SimulateAvalon.MMread(Clk, tCK, A, Rd, RQ, 1, i, None, None)

        # write a few things
        yield SimulateAvalon.MMwrite(Clk, tCK, A, WD, Wr, 0, 1)
        yield hdlutils.delayclks(Clk, tCK, 2)
[...]

Issue #15, opened 2015-02-28
toVHDL(): superfluous cast when indexing a port and numeric_ports == False


josyb:
The following code:
```python
if A[LWIDTHA_RAM]:
```
gets converted into:
```text
if bool(unsigned(A)(LWIDTHA_RAM)) then
```
The cast *unsigned( )* is unnecessary as an index operation on either an *unsigned* or  a *std_logic_vector* both return a *std_logic*
Apart from that, ModelSim flags this as an error:
```text
Vcom: Prefix of indexed name cannot be a type conversion (UNSIGNED) expression.
```

josyb:
This issue has been addressed/superseded by a commit by @jandecaluwe and follow-up PRs #80 and #87

Issue #14, opened 2015-02-28
toVHDL(): Possibly incorrect order of casting and slicing when numeric_ports == False


josyb:
The following code:
```python
@always_seq( Clk.posedge, reset = Reset)
def ccregrw():
    if Wr and (A == OFFSET):
        Q.next = WD[WIDTH+START:START]
```
gets converted to:
```vhdl
TCSR_CONTROL_CC_0_CCREGRW: process (Clk, Reset) is
begin
    if (Reset = '1') then
        control_ccassigns_6_D <= to_unsigned(0, 24);
    elsif rising_edge(Clk) then
        if ((Wr = '1') and (unsigned(A) = OFFSET)) then
            control_ccassigns_6_D <= unsigned(WD)((WIDTH + START)-1 downto START);
        end if;
    end if;
end process TCSR_CONTROL_CC_0_CCREGRW;
```
ModelSim (launched by the Sigasi editor) flags *unsigned(WD)((WIDTH + START)-1 downto START)* as an error:
```text
Vcom: Prefix of slice name cannot be type conversion (UNSIGNED)  expression.
```
Quartus and ISE (so I have been told) have no issue and compile this without a blink.
If I change it to :  unsigned(WD((WIDTH + START)-1 downto START));  ModelSim shuts up.

josyb:
This issue has been addressed/superseded by a commit by @jandecaluwe and follow-up PRs #80 and #87

PR #20, opened 2015-03-04
toVHDL(): added fix when assigning intbv()[1:] to std_logic

josyb:
This fix is related to issue #12.
In the conversion a Signal(intbv()[1:])  is assigned to a std_logic output port.
The actual code assumes that when the RHS is not a std_logic, it is a bool() and applies a stdl() conversion function, which is not defined for a Signal(intbv()[n:]) argument. 
The proposed change detects that the RHS is an intbv and supplies a subscript-index of 0 instead.

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -1,4 +1,4 @@
-
+#  This file is part of the myhdl library, a Python package for using
 #  Python as a Hardware Description Language.
 #
 #  Copyright (C) 2003-2014 Jan Decaluwe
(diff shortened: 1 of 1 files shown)


jandecaluwe:
I cherry-picked the first small commit, still have to review the second (unrelated?) one.

When you add commits in a branch, they are added to an existing pull request apparently, which is logical when you think of it. Keeping PRs separate seems to require separate branches to start them from ...

josyb:
They are unrelated.
I didn't realize that a commit was enough to be added to a PR. I wanted to discus the 'proposed fix' for #14 and #15 before issuing a PR but ended up messing this one anyway. I'll read up on branching ...

jandecaluwe:
I agree PR behavior may be counterintuitive at first. I think the way to look at it is that a PR is a view on a branch, not a fixed range of commits. Can be nice - it implies that you can fix things within a PR even after raising it, and even reopen/reuse it  (I assume) after it was merged.

I think the best is to keep work on separate issues in feature branches - this is where git excels (also over mercurial.) You can raise PRs from feature branches while keeping your own master branch in sync with the main repo. When a feature branch has been merged in, or is no longer necessary, just delete it.

josyb:
I see that you committed that first small commit. What I mostly do is to delete my fork on github and the clone on my machine and start all over. I had an occasion where my local clone wanted to commit your commits. 
So I suggest we close this PR first. The original objective has been achieved.

jck:
Next time, make sure that you start off with a branch. That way you can just delete the branch after the PR has been merged.

Also, see this regarding keeping your fork's master branch in sync: https://help.github.com/articles/syncing-a-fork/

jandecaluwe:
@josyb By using feature branches for specific work, it should no longer be necessary to delete forks and clone again all the time. You will like them :-)

PR #21, opened 2015-03-04
fix Issue 19

### LINKED ISSUE #19: conversion error when tristate signal is not read in the module.
Below is a modified version of a tristate signal example from https://gist.github.com/cfelton/6119313/download#

```python
from myhdl import *

def top(sda, scl, sda_i, sda_o, scl_i, scl_o):
     """Simple I2C bi-dir converter.
     This example will break the I2C bi-directional signals into
     uni-dir explicit signals.
     """
     sda_d = sda.driver()
     scl_d = scl.driver()

     @always_comb
     def hdl():
         #sda_i.next = sda
         sda_d.next = False if not sda_o else None

         scl_i.next = scl
         scl_d.next = False if not scl_o else None

     return hdl


def convert():
     clk = Signal(False)
     rst = Signal(False)
     sda = TristateSignal(True)
     scl = TristateSignal(True)
     sda_i = Signal(False)
     sda_o = Signal(False)
     scl_i = Signal(False)
     scl_o = Signal(False)

     toVerilog(top, sda, scl, sda_i, sda_o, scl_i, scl_o)
     toVHDL(top, sda, scl, sda_i, sda_o, scl_i, scl_o)

if __name__ == '__main__':
     convert()
```

If I comment the line `sda_i.next = sda`, the conversion result becomes incorrect.
The sda port is reported as not used, and mistakenly converted to input port. 
The assignment statement for sda is also missing.

```verilog
module top (
    sda,
    scl,
    sda_i,
    sda_o,
    scl_i,
    scl_o
);
// Simple I2C bi-dir converter.
// This example will break the I2C bi-directional signals into
// uni-dir explicit signals.

input sda;
output scl;
wire scl;
input sda_i;
input sda_o;
output scl_i;
wire scl_i;
input scl_o;

wire scl_d;
wire sda_d;

assign scl = scl_d;

assign sda_d = (!sda_o) ? 1'b0 : 'bz;
assign scl_i = scl;
assign scl_d = (!scl_o) ? 1'b0 : 'bz;

endmodule
```

The test case "general/test_ShadowSignal.py" does not capture this problem because there is not port in that testbench.

hashhsah:
fixes #19. all test cases and my local projects are passing.

One caveat remains when the tristate signal is in an interface: a local name of the interface signals must be declared in the generator function. Otherwise, _HierExtr.extract() will not add them to symdict and conversion will fail.

### DIFF
--- a/myhdl/_Cosimulation.py
+++ b/myhdl/_Cosimulation.py
@@ -138,7 +138,10 @@ class Cosimulation(object):
                     if next >= (1 << (s._nrbits-1)):
                         next |= (-1 << s._nrbits)
             except ValueError:
-                next = intbv(0)
+                if v in 'xX':
+                    next = s._init
+                else:
+                    next = intbv(0)
             s.next = next
                  
         self._getMode = 0
--- a/myhdl/_Cosimulation.py
+++ b/myhdl/_Cosimulation.py
@@ -132,15 +132,17 @@ class Cosimulation(object):
         e = buf.split()
         for i in range(1, len(e), 2):
             s, v = self._toSigDict[e[i]], e[i+1]
-            try:
-                next = int(v, 16)
-                if s._nrbits and s._min is not None and s._min < 0:
-                    if next >= (1 << (s._nrbits-1)):
-                        next |= (-1 << s._nrbits)
-            except ValueError:
-                if v in 'xX':
-                    next = s._init
-                else:
+            if v in 'zZ':
+                next = None
+            elif v in 'xX':
+                next = s._init
+            else:
+                try:
+                    next = int(v, 16)
+                    if s._nrbits and s._min is not None and s._min < 0:
+                        if next >= (1 << (s._nrbits-1)):
+                            next |= (-1 << s._nrbits)
+                except ValueError:
                     next = intbv(0)
             s.next = next
                  
--- a/myhdl/_ShadowSignal.py
+++ b/myhdl/_ShadowSignal.py
@@ -245,13 +245,15 @@ class _TristateSignal(_ShadowSignal):
     def toVerilog(self):
         lines = []
         for d in self._drivers:
-            lines.append("assign %s = %s;" % (self._name, d._name))
+            if d._driven:
+                lines.append("assign %s = %s;" % (self._name, d._name))
         return "\n".join(lines)
 
     def toVHDL(self):
         lines = []
         for d in self._drivers:
-            lines.append("%s <= %s;" % (self._name, d._name))
+            if d._driven:
+                lines.append("%s <= %s;" % (self._name, d._name))
         return "\n".join(lines)
 
 
--- a/myhdl/conversion/_analyze.py
+++ b/myhdl/conversion/_analyze.py
@@ -42,7 +42,7 @@ from myhdl.conversion._misc import (_error, _access, _kind,
                                     _ConversionMixin, _Label, _genUniqueSuffix)
 from myhdl._extractHierarchy import _isMem, _getMemInfo, _UserCode
 from myhdl._Signal import _Signal, _WaiterList
-from myhdl._ShadowSignal import _ShadowSignal, _SliceSignal
+from myhdl._ShadowSignal import _ShadowSignal, _SliceSignal, _TristateDriver
 from myhdl._util import _isTupleOfInts, _dedent, _flatten, _makeAST
 from myhdl._resolverefs import _AttrRefTransformer
 from myhdl._compat import builtins, integer_types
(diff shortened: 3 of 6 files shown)


myhdl/_Cosimulation.py
```diff
@@ -137,7 +137,10 @@ def _get(self):
                     if next >= (1 << (s._nrbits-1)):
                         next |= (-1 << s._nrbits)
             except ValueError:
-                next = intbv(0)
+                if v in 'xX':
+                    next = s._init
```
cfelton: I might have missed this in a conversation, what is the thinking behind this change?

hashhsah: In Co-simulation, icarus may return 'x', corresponding to None in MyHDL. Without the change, 'x' will be translated to intbv(0). Tristate bus relies on the rule that all but one driver's value are None, so assigning zero values to it does not work. My first attempt was:

```python
if v in 'xX':
  next = None
else:
  next = intbv(0)    # probably deadcode
```

However, the logical and arithmetic operators defined for `_Signal` do not support None value. The above code causes many test cases to fail. Not wanting to code "None support" for all operators, I'm using the fact that `_TristateSignal._init == None` while other types of signals usually have a defined initial value. This way, all previous working code hopefully will continue to work, and tristate bus correctly receives None values as it should.

This is clearly not perfect solution. I think `_Signal` operators should support None values, but being new to MyHDL, I'm not sure if this is the way to go.

cfelton: Why `x` instead of `z`?  If `z` was used for tri-stated value in cosimulation, the first approach would probably work.  I think it is an issue because all the Verilog signals are 'x' at T0, it tries to sync everything to None.  If 'z' is used instead then the None could be used.  

I am worried that for non-tristate signals, in the past an 'x' would goto 0 and now it will go back to the initial value (grant it 0 is a common init value).  This might change the behavior or existing code.

hashhsah: I see. Will test more and raise PR again.

hashhsah: Suppose I have a single tri-state output buffer on a bus. Its output enable is initially off, so the bus's value will remain at 'X'. Only after OE turned on for once and then turned off, the bus's value will change to 'Z'.

Essentially we are mapping 4-value (01XZ) logic to a 3-value (01None) logic. I think Z state is more important than 'X', because it's required in tri-state signal. As for 'X' state, assigning any of '01Z' to it is not technically wrong.

In short, the mapping comes to my mind is:
```
0 -> 0
1 -> 1
Z -> None
X -> s._init    # None for tristate, 0/1 for others.
```

myhdl/_ShadowSignal.py
```diff
@@ -244,13 +244,15 @@ def _resolve(self):
     def toVerilog(self):
         lines = []
         for d in self._drivers:
-            lines.append("assign %s = %s;" % (self._name, d._name))
+            if d._driven:
+                lines.append("assign %s = %s;" % (self._name, d._name))
```
cfelton: Not sure I 100% understand these changes.  Do you have an example what this mod is fixing?  Is it covered in the test you created?

myhdl/_Cosimulation.py
```diff
@@ -131,13 +131,18 @@ def _get(self):
...
         for i in range(1, len(e), 2):
             s, v = self._toSigDict[e[i]], e[i+1]
-            try:
-                next = int(v, 16)
-                if s._nrbits and s._min is not None and s._min < 0:
-                    if next >= (1 << (s._nrbits-1)):
-                        next |= (-1 << s._nrbits)
-            except ValueError:
-                next = intbv(0)
+            if v in 'zZ':
```
cfelton: Thanks for the explanation (in the previous commit).  I think this makes sense and seems to be a reasonable solution.

cfelton:
@hashhsah Thanks for this work, it will be nice to have some of these issues resolved with tri-states.

myhdl/_ShadowSignal.py
hashhsah: If a driver is declared with `dSig = sig.driver();`, but no logic writes to `dSig`, `dSig._driven` would be `False`. It is not necessary to generate the verilog code `assign sig = dSig;` in this case. That is, if `_driven` is reliable.

My testcase does not cover it, but I've tried this locally by adding the `Y_dummy` line to `test_tristate.py`:
```python
    Y_d = Y.driver()
    Y_dummy = Y.driver()
```

cfelton: @hashhsah perfect, that is what I assumed but I didn't want to assume :)

jck:
@hashhsah please using rebase instead of merge for keeping PR in sync.

hashhsah:
@jck . rebased.

rebase produces cleaner commit logs, but re-writes some history. For private branches, I always use rebase. For branches already published in a PR, I don't have a definitive answer, but slightly lean toward merge for history preservation.

jck:
I'm also not 100% sure, but we should come up with a consensus/best practices.

I'd like to create a CONTRIBUTING.md and update http://dev.myhdl.org/guide.html

cfelton:
@jck hmmm, don't like the dual maintenance, maybe README.md should have a contributing section and refer to the the guide

jck:
@cfelton CONTRIBUTING could just have a link to the dev guide. The good part about having that file is that github prompts users to read it when they open a PR.

https://github.com/blog/1184-contributing-guidelines

cfelton:
@jck, makes sense, I agree with the CONTRIBUTING and have it point to the dev guide.

PR #22, opened 2015-03-09
Fix GHDL test, run GHDL tests in travis

jck:


jandecaluwe:
Quite some "details" no in ci.sh, can we factor these out?

cfelton:
I agree with factoring `ci.sh`, also think if we are going to have multiple scripts we should create `scripts/ci`.

PR #26, opened 2015-03-11
simplify ci.sh

jck:
This commit brings ci.sh down to 36 lines by removing the coloring and heading functions.


jandecaluwe:
PRs are great, one just specifies what one wants and gets a PR the next morning ;-)

Issue #33, opened 2015-03-16
Constants are not correctly handled when converting code with attribute references


jck:
https://gist.github.com/hgomersall/4b3d0f01a14863551007

PR #37, opened 2015-03-18
Test constant in interfaces

cfelton:
Added a test to the existing interface tests to verify constants embedded in the interfaces.  Currently this works with Verilog conversion but not with VHDL conversion.  This pull-request will fail the CI test, this is expected until a fix is put in place.  This error has been reported in issue #33.


myhdl/test/conversion/general/test_interfaces3.py
```diff
@@ -100,17 +171,36 @@ def test_one_verify():
...
 def test_two_verify():
     assert verify(c_testbench_two) == 0
 
+def test_three_analyze():
+    clock = Signal(bool(0))
+    reset = ResetSignal(0, active=0, async=True)
+    x = Signal(intbv(3, min=-5000, max=5000))
+    y = Signal(intbv(4, min=-200, max=200))
+    intf = IntfWithConstant2()
+    analyze(m_top_const, clock, reset, x, y, const)
```
jck: should this be intf.const ?

cfelton: Yes, I agree.  now wondering why it worked?

jck:
I have an idea brewing which will allow us to keep this test without disrupting travis using pytest marks.
@cfelton I'll send a PR to your branch when I'm done.

cfelton:
@jck that is a nice little trick to suppress failing tests.  But I think if we find a bug and we create a test for the bug we really should be showing **failing**?  What do you think @jandecaluwe?  Is your policy only to merge PRs that are passing the all tests?

Of course this gets complicated, what do we expect *master* to be?  If it is development, it isn't unheard of that tests would be failing, especially if we use a TDD approach.  But it is also the landing page where many new users first visit; a green light is more inviting than a red. 

@jck do you know if there a method to get a summary back from travis?  Could we categories released features vs. features being developed (green and red)?  I haven't seen a method to do this.

jck:
In this case, we don't really know when this bug will be fixed. Xfail logs
that there has been an expected failure but does not mark the whole build
as a failure. I think this is important because if master is indefinitely
in a failing state according to Travis, then we can't tell by glancing at a
new PR if it has created any more bugs than the already existing ones.

cfelton:
> if master is indefinitely in a failing state according 
> to Travis, then we can't tell by glancing at a new 
> PR if it has created any more bugs than the already 
> existing ones

I agree with the point and that is probably reason enough.  But I am still troubled that we might "forget" about the failing tests (out of sight, out of mind).  Locally it does give an indication that the test is failing.  In the end, agree with you, no point in crippling the convenient PR checks, we will need to be vigilant creating issues (#33) and checking the local test.

Thanks!

jandecaluwe:
I guess this is an exception - normally we would want all tests to pass in a PR. I understand the value here though, and I think the proposed temporary fix is OK.

cfelton:
Often it is easier (faster) to create a new test than 
apply the fix, the test will often come before the fix :)
 
Maybe if we apply a label to a PR ("new test" or something) 
the PR won't be merged until the fix is merged.  That way we 
still know there is an issue in the master by looking at the 
PRs with the label.  And possibly we want to encourage a
test-PR when an issue is created.  The downside the developer
will need to merge with the test-PR vs. master.

But in general the test-PR will be abandoned / deleted because 
the fix-PR will need the test incorporated / merged (chicken 
and egg) or I guess if the test-PR is merged it will be 
benign because the changeset was added in the fix-PR?

I do like the idea of a test-PR with all reported 
"bug issues".  If there are a bunch of failing test-PRs
we know we have an issue.

jck:
I propose that test-PRs use the decorator I introduced in #45. New test PRs can be merged immediately. When someone implements a fix for a bug, they can delete the line in their PR.

jandecaluwe:
OK

PR #39, opened 2015-03-23
VCD dump of tristate signals

hashhsah:
Due to pull request #21, which fixes issue #19, TristateSignals may contain 'None' value when we dump it to .vcd file. This PR writes 'z' to .vcd if it sees 'None' value.

### DIFF
--- a/myhdl/_traceSignals.py
+++ b/myhdl/_traceSignals.py
@@ -35,6 +35,7 @@ import shutil
 from myhdl import _simulator, __version__, EnumItemType
 from myhdl._extractHierarchy import _HierExtr
 from myhdl import TraceSignalsError
+from myhdl._ShadowSignal import _TristateSignal, _TristateDriver
 
 _tracing = 0
 _profileFunc = None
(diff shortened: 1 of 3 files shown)


hashhsah:
I don't have a vcd standard document at hand, and only tested the generated .vcd with gtkwave. More tests and comments please.

myhdl/_traceSignals.py
```diff
@@ -148,7 +149,14 @@ def _writeVcdSigs(f, hierarchy, tracelists):
                 print("$upscope $end", file=f)
         print("$scope module %s $end" % name, file=f)
         for n, s in sigdict.items():
-            if s._val is None:
+            if isinstance(s, _TristateSignal):
```
hashhsah: The original code is mostly agnostic about implementation details in Signal, the new code relies on more such details. examples: 
 - with `isinstance()`, we must import `_TristateSignal` and `_TristateDriver`
 - using `s._orival` in addition to `s._val`

Ugly. But I don't see any simple way around it.

jandecaluwe:
I have restarted Travis a few times on this, and apparently the build works, but still the message from Travis here doesn't say so. Does anyone have a clue?

hashhsah:
When I raised this PR, the travis system was a little sluggish. Builds of some other of my branches hung for over ten minutes before starting to build. Perhaps a rare bug in either github or travis was triggered then.

I can push a dummy commit, or raise this PR again, if necessary.

jck:
I've reported the issue to travis.

Issue #40, opened 2015-03-24
Conversion failure if integer comparison used in if-else


cfelton:
The following failing code example was presented on the mailing-list by @geschema.

```python
from myhdl import *

def mpegChannel(clk, rst):

    s_tx_data_xor_mask_r = Signal(intbv(0)[1 + 31:])

    @always_seq(clk.posedge, rst)
    def fsm_seq():
        for i in range(4):
            if i == 0:
                s_tx_data_xor_mask_r.next[1 + 7:0] = 0
            elif i == 1:
                s_tx_data_xor_mask_r.next[1 + 15:8] = 1
            elif i == 2:
                s_tx_data_xor_mask_r.next[1 + 23:16] = 2
            else:
                s_tx_data_xor_mask_r.next[1 + 31:24] = 3

    return instances()

clk = Signal(bool(0))
rst = ResetSignal(0, active=1, async=True)
mpegChannelInst = toVerilog(mpegChannel, clk, rst)
```

The above code gives a type error in during analysis.

```python
myhdl\conversion\_analyze.pyc in visit_If(self, node)
    747         node.caseVar = var1
    748         node.caseItem = item1
--> 749         if (len(choices) == _getNritems(var1.obj)) or node.else_:
    750             node.isFullCase = True

myhdl\conversion\_analyze.pyc in _getNritems(obj)
    401         return len(obj._type)
    402     else:
--> 403         raise TypeError("Unexpected type")
    404 
    405 

TypeError: Unexpected type
```

A solution was proposed by @josyb .  The solution is two parts, first to short-circuit on the `else` and
then to identify the `int,long` in the VHDL conversion.

cfelton:
A fix has been merged, PR #41.

cfelton:
@jandecaluwe I believe this issue can be closed.  @geschema have you merged the latest and verified it fixes the issue you observed?

PR #42, opened 2015-03-25
Issue #40 test.

cfelton:
This is a test for issue #40.  Issue #40 fails to convert when an "integer" is compared in a full-case if-else.

This test will fail travis-ci until a fix is merged or `xfail` is used to mute the test.  At this point, I didn't include `xfail`.  This PR can be merged after a fix is in place.

### DIFF
--- a/myhdl/test/bugs/test_github_issue_44.py
+++ b/myhdl/test/bugs/test_github_issue_44.py
@@ -1,0 +1,82 @@
+from __future__ import absolute_import
+
+from random import randint
+
+from myhdl import *
+from myhdl.conversion import verify
+
+def m_int_case(clock, reset, x, y):
+ 
+    @always_seq(clock.posedge, reset=reset)
+    def rtl():
+        z = (x >> 16) & 0x3
+        if z == 0:
+            y.next = y + 0xDECAFBAD
+        elif z == 1:
+            y.next = x + 0xC0FFEE
+        elif z == 2:
+            y.next = y + 2
+        else:
+            y.next = 299792458
+
+    return rtl
+
+ 
+def c_testbench():
+    """ convertible testbench """
+    clock = Signal(bool(0))
+    reset = ResetSignal(0, active=0, async=False)
+    x = Signal(intbv(0, min=0, max=444356))
+    y = Signal(intbv(0)[32:])
+
+    tbdut = m_int_case(clock, reset, x, y)
+
+    @instance
+    def tbclk():
+        clock.next = False
+        while True:
+            yield delay(3)
+            clock.next = not clock
+
+    @instance
+    def tbstim():
+        reset.next = reset.active
+        yield delay(11)
+        reset.next = not reset.active
+
+        # this is a converion issue, test a couple cases to
+        # make sure the module is accurate
+        
+        # value after reset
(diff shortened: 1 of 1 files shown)


josyb:
@cfelton I tested my fix against your test and it passes
```text

C:\Python27\Lib\site-packages\myhdl\test\bugs> py.test GHDL.py .\test_bug_44.py
============================= test session starts =============================
platform win32 -- Python 2.7.9 -- py-1.4.20 -- pytest-2.5.2
collected 1 items

test_bug_44.py .

========================== 1 passed in 0.14 seconds ===========================
```
Travis CI is **lagging** for several hours?

cfelton:
@josyb looks like it completed successfully.

jandecaluwe:
This test was broken in a number of ways - removed it. Better to stay as close as possible to the original test case, which was fine as a test.

josyb:
Yes, I got tricked by the _44 
The intended test, test_github_issue_44.py, failed:
```text
C:\Python27\Lib\site-packages\myhdl\test\bugs> py.test GHDL.py .\test_github_issue_44.py
============================= test session starts =============================
platform win32 -- Python 2.7.9 -- py-1.4.20 -- pytest-2.5.2
collected 2 items

test_github_issue_44.py .F

================================== FAILURES ===================================
____________________________ test_github_issue_44 _____________________________

    def test_github_issue_44():
>       assert verify(c_testbench) == 0
E       assert 1 == 0
E        +  where 1 = verify(c_testbench)

test_github_issue_44.py:79: AssertionError
------------------------------- Captured stderr -------------------------------
pck_myhdl_09.vhd:179:14:warning: function "normalise" is never referenced
c_testbench.vhd:20:15: ',' or ':' is expected after identifier in constant declaration
Analysis failed
===================== 1 failed, 1 passed in 0.39 seconds ======================
```
because of this:
```python
    @instance
    def tbstim():
        reset.next = reset.active
        yield delay(11)
        reset.next = not reset.active
```
resulting in:
```VHDL
constant reset.active: integer := 0;
```
Anyway, I pulled in the new test and verified it passed OK (in my local development tree).

jandecaluwe:
That was just the first issue - there were several others and I didn't bother to fix them. In general, I think the test should be as close a possible to the originally reported issue (trivial in this case), that it should guide the fix and come as part of the PR with the fix.

cfelton:
@jandecaluwe thanks for fixing the test.

To be fair, the issues encountered, more than likely, would have occurred with the original example.  The issues were in the convertible testbench not the updated example.

There were questions/concerns/protests with the original example, mainly because the example could simply be replaced with an assignment:

```python
    long_signal_name.next = 0x03020100
```

It seemed distracting.  I tried to create an example and test that couldn't be trivially reduced but not too complex to obscure the issue.

For future reference, my mistakes were common testbench conversion pitfalls:

   1. myhdl.conversion.verify require prints in the convertible testbench (I frequently forget these).

   2. Initial value cosimulation mismatch, I didn't catch this because the example didn't convert.

   3. Failure to convert attribute constants, VHDL only.  Even though this is an outstanding issue, my mistake was basically the same as the above, me not being judicious with convertible testbenchs.

PR #47, opened 2015-03-29
travis: use container based infrastructure

jck:
This has various benifits. Most notably, builds start considerably
faster and perform slightly better.


jck:
It looks like this doesn't work on PRs yet. I'll leave it open and update when it's ready.

jck:
This is ready now. I've also added testing of bugs and improved the modelsim makefile.

PR #49, opened 2015-04-03
Assign initial zero values to standard data types

masterdezign:
Initializing signals at least with zero values is considered a good practice in C/C++. My experience points the same truth is applicable to VHDL.

See also
 http://stackoverflow.com/questions/6363130/is-there-a-reason-to-initialize-not-reset-signals-in-vhdl-and-verilog
 http://stackoverflow.com/questions/16814653/vhdl-signal-initialisation

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -387,9 +387,13 @@ def _writeSigDecls(f, intf, siglist, memlist):
                 warnings.warn("%s: %s" % (_error.UnreadSignal, s._name),
                               category=ToVHDLWarning
                               )
-            # the following line implements initial value assignments
-            # print >> f, "%s %s%s = %s;" % (s._driven, r, s._name, int(s._val))
-            print("signal %s: %s%s;" % (s._name, p, r), file=f)
+            # Initial zero value assignments
+            if p in ['signed', 'unsigned', 'std_logic_vector']:
+                print("signal %s: %s%s := (others => '0');" % (s._name, p, r), file=f)
+            elif p == 'std_logic':
+                print("signal %s: %s%s := '0';" % (s._name, p, r), file=f)
+            else:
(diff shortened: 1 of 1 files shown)


jmgc:
Dear all,

In my humble opinion, I think the "zeroing" of signals must be done during the reset, in order to ensure post-synthesis equivalent behaviours. If not, debugging it becomes quite painful, because U or X start to appear everywhere.

Best regards

cfelton:
To have consistent Cosimulation results it is desired to set the V* initial values the same as the MyHDL initial values.  The initial values should not be set to zero but set to the value indicated at the instance creation.  

Initial value support exists in the code, Initial value support can be enabled but there was a problem in the past with inconsistent behavior with synthesis tools (e.g XST didn't support Verilog initial value support).  The testing needs to be completed and recorded.

If some designers prefer no initial value, I believe it was discussed in the past that `None` could be used ... but this overlaps with the `None` usage for 'z'.  This would need to be revisited (discussed).

See:
http://dev.myhdl.org/tasks/initial-values-support.html
http://dev.myhdl.org/ideas/initial-values.html

masterdezign:
MyHDL is a democratic tool. I believe, in democracy, the designer should have a choice to use initialisation or not.

Here is a simple example when initialisation might be preferable: a simple modulation (you may call it a clock divider if you wish).

```
    @always_seq(clk.posedge, reset=reset)
    def mod():
        if counter == SOME_CONST:
            new_clk.next = not new_clk
            counter.next = 0
        else:
            counter.next = counter + 1

```

Here both counter and new_clk have to be initialised. 

@jmgc please point me to a better realisation if the above does not meet optimal design requirement.

Thank you,
B

masterdezign:
@cfelton http://dev.myhdl.org/tasks/initial-values-support.html 
That would be perfect.

jandecaluwe:
Closing following the discussion.

Issue #27, opened 2015-03-11
make icarus fails in test/bugs if ghdl is not installed.


jck:
https://travis-ci.org/jck/myhdl/jobs/53907805#L400

udara28:
I found that test fails when trying conversion.analyz, to fix this I added a new line to icarus.py which set analyze.simulator to icarus. I submitted a patch <a href="https://github.com/jandecaluwe/myhdl/pull/50">https://github.com/jandecaluwe/myhdl/pull/50</a> as a fix for this issue.

jandecaluwe:
Cherry-picked relevant commit from #50.

PR #50, opened 2015-04-04
Patch issue 27

udara28:
This patch is to fix the issue <a href="https://github.com/jandecaluwe/myhdl/issues/27">https://github.com/jandecaluwe/myhdl/issues/27</a>
When trying to 'make icarus' some tests will fail if ghdl is not installed. 

The failuture is due to icarus.py only setting verify.simulator to icarus which cause conversion.analyze try to use ghdl. In the fix, Added a line to icarus.py so that analyze.simulator is also set to icarus.

Even though it is not mandatory to do the same for GHDL.py, same change is added to GHDL.py to avoid confusions in future.


jandecaluwe:
Cherry-picked only the relevant commit.

PR #53, opened 2015-04-05
Adding tests for sigsIn2DLists

ravijain056:
Added a simple test to verify that the conversion of 2d los fails according to the current codebase.....


jandecaluwe:
Too complicated, test should be same for Verilog/VHDL, and fail with a specific kind of error.

PR #56, opened 2015-04-07
test_always_seq.py

iamsrinivas:
Added new test to increase the coverage of the code. The tests do  not reach the full coverage(100%). We need to extend this test to improve the coverage hence forward using this test.

### DIFF
--- a/myhdl/test/core2/test_always_seq.py
+++ b/myhdl/test/core2/test_always_seq.py
@@ -28,7 +28,7 @@ def test_clock():
         def logic2():
             pass
     except:
-        pytest.fail()
+        assert False
 
 def test_reset(): 
     """ check the reset parameter """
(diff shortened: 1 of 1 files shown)


myhdl/test/core2/test_always_seq.py
```diff
@@ -0,0 +1,54 @@
...
+        assert e.kind == _error.EdgeType
+
+    # should work with a valid Signal
+    clock = Signal(bool(0))
+    try:
+        @always_seq(clock.posedge, reset=reset)
+        def logic2():
+            pass
+    except:
+        pytest.fail()
```
jandecaluwe: How can this line work if `pytest` is not imported? Why not `assert False` as a generic statement?

myhdl/test/core2/test_always_seq.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    # should fail without a valid ResetSignal
+    clock = Signal(bool(0))
+    reset = Signal(bool(0))
+
+    with raises(AlwaysSeqError):
+        @always_seq(clock.posedge, reset=reset)
+        def logic():
+            pass
+        assert e.kind == _error.ResetType
```
jandecaluwe: How can this work if `e` is not defined?

PR #57, opened 2015-04-09
add test targets for TriState

Mucka:
- Simple test for TriState moved from test_ShadowSignal to test_TriState
- Slicing and concatenating tests for TriState added to test_TriState
- 3 discovered issues (not sure if it's lack of my understanding or real issue) added to test/bugs/test_bug_tristate with explanation in comments

### DIFF
--- a/myhdl/test/bugs/test_bug_tristate.py
+++ b/myhdl/test/bugs/test_bug_tristate.py
@@ -1,0 +1,81 @@
+from __future__ import absolute_import
+from myhdl import *
+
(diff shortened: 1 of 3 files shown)


hashhsah:
Is this a bug report? I did some work recently on tristate in myhdl. Could you raise an issue to explain your failing test case a little?

jandecaluwe:
Possible bugs should be reported as such, not as a PR that fails the regression suite.

PR #58, opened 2015-04-09
Add support to transport delays on DelayedSignals as supported in VHDL.

rodrigosurita:
Just an idea that worked fine for me while simulating propagation delays. Didn't implement VHDL/Verilog conversion yet.


jandecaluwe:
No test and obviously wasn't even tested locally.

PR #63, opened 2015-04-14
Various cosimulation and testing improvements.

jck:


cfelton:
I like the changes, combining the core tests into a single directory and using py.test as the test runner.

hashhsah:
+1 on switching to subprocess.

Is there a fundamental problem preventing co-simulation to work on pypy?

jck:
So far, we were not running test_Cosimulation.py through test_all.py.
Cosimulation actually works fine on pypy, but this test does not. I believe it is because this test depends on the behaviour of __del__, which is implementation specific.

PR #64, opened 2015-04-16
Rename attribute references in the AST before analysis

### LINKED ISSUE #33: Constants are not correctly handled when converting code with attribute references

jck:
fixes #33

### DIFF
--- a/myhdl/_resolverefs.py
+++ b/myhdl/_resolverefs.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import ast
+import itertools
 from types import FunctionType
 
 from myhdl._util import _flatten, _makeAST, _genfunc
@@ -24,6 +25,10 @@ def _resolveRefs(symdict, arg):
 
 #TODO: Refactor this into two separate nodetransformers, since _resolveRefs
 #needs only the names, not the objects
+
+def _suffixer(name):
+    suffixed_names = (name+'_renamed{0}'.format(i) for i in itertools.count())
+    return itertools.chain([name], suffixed_names)
 
 
 class _AttrRefTransformer(ast.NodeTransformer):
@@ -52,13 +57,12 @@ class _AttrRefTransformer(ast.NodeTransformer):
                 return node
 
         attrobj = getattr(obj, node.attr)
-        new_name = node.value.id+'.'+node.attr
-        if new_name not in self.data.symdict:
-            self.data.symdict[new_name] = attrobj
-            self.data.objlist.append(new_name)
-        else:
-            pass
-            #assert self.data.symdict[new_name] == attrobj
+
+        name = node.value.id + '_' + node.attr
+        new_name = next(s for s in _suffixer(name) if s not in self.data.symdict)
+        self.data.symdict[new_name] = attrobj
+        self.data.objlist.append(new_name)
+
         new_node = ast.Name(id=new_name, ctx=node.value.ctx)
         return ast.copy_location(new_node, node)
 
--- a/myhdl/conversion/_analyze.py
+++ b/myhdl/conversion/_analyze.py
@@ -57,15 +57,6 @@ _extConstDict = {}
 
 
 def _makeName(n, prefixes, namedict):
-    #Take care of names with periods
-    #For attribute references, periods are replaced with '_'.
-    if '.' in n:
-        n = n.replace('.', '_')
-        if n in namedict:
-            i = 0
-            while (n + '_{0}'.format(i)) in namedict:
-                i += 1
-            n += '_{0}'.format(i)
     # trim empty prefixes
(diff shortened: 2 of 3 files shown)


jmgc:
Dear @jck,

I have merged this pull request, and I have found a new bug.

A 'Signal has multiple drivers' error is generated in a code that worked perfectly well before the pull was done. The reason seems to be that although the signals generated by the present pull, have fully different names including the '_renamed{n}', they share the same signal, which stores the '._driven' attribute. As a result, both signals are seen as the same one in terms of the '._driven' attribute, though, in principle, they should be considered different ones (I suppose).
I am trying to prepare some code to show detect the bug.
Best,
Jose M.

jck:
@jmgc I'll look into it. Please share a failing example.

jmgc:
Dear @jck,

I made the tests using the numeric add-on that I have made. I have to convert the code to intbv to confirm the issue.

Sorry for not answering before.

jmgc:
Dear @jck,

Yesterday I had sometime to revise a bunch of bugs that I had in my branch of myhdl. While I was doing this, also had a look on the changes you made, and checked one small thing.

I applied part of your code: removed the code from the _analyze file and substituted the '.' by a '_' in the _resolvrefs (name = node.value.id + '_' + node.attr), and removed the bug on test_interface3.py
I also modified all the interfaces in the tests to become child of object.

The result was that all the tests were passed but the test_interface2.py related with having two variables with the same name a_x.

I do not know if this can be helpful. In any case, I will try to prepare the bug file in the next days.

jmgc:
Dear @jck,

You can find the code in the gist: https://gist.github.com/0fb34c528b9ab59a0256.git

It is a big large, but I wanted to provide you the full code so you can have a better view of the effect.

The resulting error is:

======================================================================
ERROR: test_DelayImageSimulate (build.test_pull_64.TestDelayImage)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/chema/Documents/Workspace/Sophi/ISSControl/myhdl/build/test_pull_64.py", line 489, in test_DelayImageSimulate
    self.assertEqual(conversion.verify(delayImageTestBench), 0)
  File "/Users/chema/Documents/Workspace/myhdl-git/myhdl/conversion/_verify.py", line 128, in __call__
    inst = toVHDL(func, *args, **kwargs)
  File "/Users/chema/Documents/Workspace/myhdl-git/myhdl/conversion/_toVHDL.py", line 168, in __call__
    genlist = _analyzeGens(arglist, h.absnames)
  File "/Users/chema/Documents/Workspace/myhdl-git/myhdl/conversion/_analyze.py", line 185, in _analyzeGens
    v.visit(tree)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/ast.py", line 245, in visit
    return visitor(node)
  File "/Users/chema/Documents/Workspace/myhdl-git/myhdl/conversion/_analyze.py", line 1070, in visit_Module
    self.raiseError(node, _error.SigMultipleDriven, n)
  File "/Users/chema/Documents/Workspace/myhdl-git/myhdl/conversion/_misc.py", line 149, in raiseError
    raise ConversionError(kind, msg, info)
myhdl.ConversionError: in file /Users/chema/Documents/Workspace/Sophi/ISSControl/myhdl/build/test_pull_64.py, line 375:
    Signal has multiple drivers: inp_image_full_renamed4

Best

jmgc:
Dear @jck,

I've been making some tests, and I think the implications of the present pull are larger, at least from my point of view.

I will talk about VHDL as is where I have some experience. The first solution I have tried is just substitute '.' by '_Dot_' instead of using the prefixer. This works quite well, and from the python point of view I think it is a convenient solution. First because it follows no python rules for naming, so no one should use it, and also, because it allows to achieve the main goal of the pull, without the problems detected previously.

The problem with this solution is that it is based on the casing of the characters, while in VHDL there is no casing. As a result, the solution is only valid if the user does not use '_dot_' p.e. It also implies that if two variables have the same name but with different casing (for example, because one is a 'constant'), they will collide in VHDL.

As a result, I think your idea of using escaped and extended identifiers should be revisited.

jmgc:
Sorry, I forgot to add a reason why I think the prefixer should be avoided, it is mainly because it could lead to a change in the name of the ports on every conversion, which can become a burden.

josyb:
While working on a fix for the detected issue in PR #80: ShadowSignals from ports are not renamed, I ended up with one test from conversion/general failing: name_conflict_after_replace.vhd
As I moved the _analyzeTopFunc forward in the conversion the _suffixer() code is skipped ...
But IMHO I think it is better to raise an error in MyHDL than to resolve such conflict silently.

PR #65, opened 2015-04-16
test_always_seq

iamsrinivas:
Fixed bugs in 56.


cfelton:
@iamsrinivas you didn't need to create a new pull-request.  You should of been able to update #56.  This is preferred because the changes are easily diffed.  If you make the changes on the original branch `iamsrinivas-patch-2` and push to you github the original #56 would be updated.

PR #67, opened 2015-05-01
A new simulation instance clears the signal state.

hgomersall:
Consider this [gist](https://gist.github.com/hgomersall/3aa0fe0eaec408555a7d).

There is a distinct problem in which the two simulators step on each other due to the signal state from the first sim stepping on the toes of the second sim.

This PR clears the old simulation state when a new simulation instance is created.

### DIFF
--- a/myhdl/_Simulation.py
+++ b/myhdl/_Simulation.py
@@ -70,6 +70,11 @@ class Simulation(object):
         self._finished = False
         del _futureEvents[:]
         del _siglist[:]
+
+        # We need to clear all the signals since we have a new simulation
+        # instance. This will break other simulation instances.
+        for s in _signals:
+            s._clear()
         
         
     def _finalize(self):
--- a/myhdl/test/core/test_Simulation.py
+++ b/myhdl/test/core/test_Simulation.py
@@ -29,6 +29,7 @@ random.seed(1) # random, but deterministic
 
 from myhdl import Simulation, SimulationError, now, delay, StopSimulation, join
 from myhdl import Signal, intbv
+from myhdl import instance, always
 from myhdl._Simulation import _error
 
 from myhdl._simulator import _siglist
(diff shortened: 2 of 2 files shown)


jandecaluwe:
If you run a simulation for a certain duration, it is explicitly the intention that you can resume it safely. If a new simulation instance would clear the state, that becomes impossible. Not a good idea.

Clearly it is not possible to have more than one active simulation instance. Therefore, what we can do is refuse to create a new instance if there is still an active one (to be checked directly as part of the global simulator state) and expose the `_finalize()` method in the API (e.g. as `quit()`).

hgomersall:
Yes, that would work too.

hgomersall:
I'll do a new PR tomorrow updated to be that way around.

hgomersall:
This seems to be much more work than I thought it would be. It seems most of the simulation tests don't leave the global state in it's starting state (that is, _finalize is never called).

How would you feel about a more complete effort to have a per simulation set of attributes - a sort of quasi non-global state. This would actually be how I'd begin the transition to eliminating global state, so it might be worthwhile anyway.

Essentially, there would be a dict of simulation attributes in _simulator.py, the keys of which are the hashes of the simulation instances. Each dict contains the attributes in _simulator.py (or some subset to begin with). Eventually, all the running state can be moved to that dict, including if desired the signal state which can be copied in as necessary. Each dict then only references it's own entry in the state dictionary.

I began along this path (just the _time attribute to begin with), but it wasn't clear to me how much this would break other things.

hgomersall:
With respect, the issue hasn't been solved.

hgomersall:
Though it wasn't an issue. My apologies :)

jandecaluwe:
I have closed this because I will not accept this PR as indicated. However, I do recognize that there is an issue and I have described the type of PR I will gladly accept. (I don't currently understand what is difficult about the solution I proposed.

PR #76, opened 2015-05-23
replaced call to built-in 'bin' function with call to 'bin' function from _bin.py

josyb:
The built-in *bin* function omits *leading zeroes* which results in a mismatch of the lengths between the LHS and the RHS. Using the MyHDL *bin* function with a width specifier remedies this.

### DIFF
--- a/myhdl/_ShadowSignal.py
+++ b/myhdl/_ShadowSignal.py
@@ -33,8 +33,8 @@ from myhdl._simulator import _siglist
 from myhdl._bin import bin
 
 # shadow signals
-        
-        
+
+
 class _ShadowSignal(_Signal):
 
     __slots__ = ('_waiter', )
@@ -49,7 +49,7 @@ class _ShadowSignal(_Signal):
         raise AttributeError("ShadowSignals are readonly")
 
 
-        
+
 class _SliceSignal(_ShadowSignal):
 
     __slots__ = ('_sig', '_left', '_right')
@@ -84,7 +84,7 @@ class _SliceSignal(_ShadowSignal):
             yield sig
 
     def _setName(self, hdl):
-        if self._right is None:       
+        if self._right is None:
             if hdl == 'Verilog':
                 self._name = "%s[%s]" % (self._sig._name, self._left)
             else:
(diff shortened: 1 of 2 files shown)


josyb:
In stead of relying on Travis to flag any error, I ran the tests locally. I don't have cosimulation working, yet, but at least the vlog, GHDL and vcom testing works.
I also expanded the test for bench_ConcatSignalWithConsts()

josyb:
And still Travis flagged an error, but it is not clear to me why.
```text
=================================== FAILURES ===================================
_________________________ test_ConcatSignalWithConsts __________________________
    def test_ConcatSignalWithConsts():
>       Simulation(bench_ConcatSignalWithConsts()).run()
test_ShadowSignal.py:96: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/myhdl/_Simulation.py:135: in run
    waiter.next(waiters, actives, exc)
/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/myhdl/_Waiter.py:129: in next
    clause = next(self.generator)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    @instance
    def check():
        for i in range(2**len(a)):
            for j in (0, 1):
                for k in range(2**len(c)):
                    for m in range(2**len(d)):
                        a.next = i
                        b.next = j
                        c.next = k
                        d.next = m
                        yield delay(10)
>                       assert s[20:18] == long(c1, 2)
E                       NameError: name 'long' is not defined
test_ShadowSignal.py:84: NameError
==================== 1 failed, 304 passed in 18.97 seconds =====================
make: *** [all] Error 1
make: Leaving directory `/home/travis/build/jandecaluwe/myhdl/myhdl/test/core'
[FAILED] make -C myhdl/test/core 
```
I don't have this offending line in my branch.

jmgc:
In python 3.4 long is not defined. You have to use the definition from _compat.

jandecaluwe:
Wait - all calls to long() should stay untouched, the _compat module should be used to have compatible definition between Py2K and Py3K. See other modules where it is done like that.

jandecaluwe:
This is becoming hard to follow - I'll pick the relevant commits and fix the rest.

jandecaluwe:
I have actually merged the relevant commits, but as this is an incomplete merge it doesn't automatically show. Thanks for the bug report and fixes.

josyb:
I'll remember about the long() next time. Maybe install Python 3.4 too. And get CoSimulation under Windows, nut there I will probably need some help.
Regards,
Josy

PR #79, opened 2015-05-25
Tristate numeric default value

jmgc:
As indicated in the closing of pull #75, the default value is defined in the Tristate numeric signal.

### DIFF
--- a/myhdl/_Signal.py
+++ b/myhdl/_Signal.py
@@ -32,7 +32,6 @@ from __future__ import print_function
 from inspect import currentframe, getouterframes
 from copy import copy, deepcopy
 import operator
-import warnings
 
 from myhdl._compat import integer_types, long
 from myhdl import _simulator as sim
--- a/myhdl/_Signal.py
+++ b/myhdl/_Signal.py
@@ -470,25 +470,13 @@ class _Signal(object):
     # conversions
     
     def __int__(self):
-        if self._val is None:
-            warnings.warn("Metavalue detected in conversion")
-            return int(0)
-        else:
-            return int(self._val)
+        return int(self._val)
         
     def __long__(self):
-        if self._val is None:
-            warnings.warn("Metavalue detected in conversion")
-            return long(0)
-        else:
-            return long(self._val)
+        return long(self._val)
 
     def __float__(self):
-        if self._val is None:
-            warnings.warn("Metavalue detected in conversion")
-            return 0.0
-        else:
-            return float(self._val)
+        return float(self._val)
     
     def __oct__(self):
         return oct(self._val)
@@ -497,11 +485,7 @@ class _Signal(object):
         return hex(self._val)
     
     def __index__(self):
-        if self._val is None:
-            warnings.warn("Metavalue detected in conversion")
-            return int(0)
-        else:
-            return int(self._val)
+        return int(self._val)
 
 
     # comparisons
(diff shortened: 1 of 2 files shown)


jandecaluwe:
I have to insist. The fact that a conversion to an unknown is attempted points in my view to a **real** error: something that will not work, or work unreliably in hardware. You don't want to mask those: you need to fix the design. I will need to be convinced of the opposite before accepting this.

If you do want to mask the problem, there is always the option of catching and handling it explicitly in the model.

jmgc:
I will ask you another question, why this code should not be converted?

```python
def bench_TristateSignal():
    s = TristateSignal(intbv(0)[8:])
    a = s.driver()
    b = s.driver()
    c = s.driver()

    @instance
    def check():
        a.next = None
        b.next = None
        c.next = None
        yield delay(10)
        #print s
        a.next = 1
        yield delay(10)
        print(int(s))
        a.next = None
        b.next = 122
        yield delay(10)
        print(int(s))
        b.next = None
        c.next = 233
        yield delay(10)
        print(int(s))
        c.next = None
        yield delay(10)
        #print s
    
    return check


def test_TristateSignal():
    assert conversion.verify(bench_TristateSignal) == 0
```
It’s equivalent in VHDL works perfectly well:

```vhdl
library IEEE;
use IEEE.std_logic_1164.all;
use IEEE.numeric_std.all;
use std.textio.all;

use work.pck_myhdl_090.all;

entity bench_TristateSignal is
end entity bench_TristateSignal;


architecture MyHDL of bench_TristateSignal is





signal a: unsigned(7 downto 0);
signal s: unsigned(7 downto 0);
signal b: unsigned(7 downto 0);
signal c: unsigned(7 downto 0);

begin


s <= a;
s <= b;
s <= c;


BENCH_TRISTATESIGNAL_CHECK: process is
    variable L: line;
begin
    a <= (others => 'Z');
    b <= (others => 'Z');
    c <= (others => 'Z');
    wait for 10 ns;
    a <= to_unsigned(1, 8);
    wait for 10 ns;
    write(L, to_integer(s));
    writeline(output, L);
    a <= (others => 'Z');
    b <= to_unsigned(122, 8);
    wait for 10 ns;
    write(L, to_integer(s));
    writeline(output, L);
    b <= (others => 'Z');
    c <= to_unsigned(233, 8);
    wait for 10 ns;
    write(L, to_integer(s));
    writeline(output, L);
    c <= (others => 'Z');
    wait for 10 ns;
    wait;
end process BENCH_TRISTATESIGNAL_CHECK;

end architecture MyHDL;
```

But it cannot be directly translated because it generates an error. While with this small change, everything works fine. And if a conversion is done from None, a warning is generated.

jandecaluwe:
No, the code you provided simulates and converts just fine. Your own code therefore makes my point and not whatever point you were trying to make. Enough time wasted. Rejected.

jmgc:
Sorry to be a little bit pushy, it seems that my version is becoming too different from the master one, and I should check on both.

In any case, I have found the example that I wanted to show you using the master branch.

In the first case:

```python
python -c "from myhdl import *;s = Signal(intbv(0)[8:]); print(int(s))"
```

The result is 0.

While if a TristateSignal is used:

```python
python -c "from myhdl import *;s = TristateSignal(intbv(0)[8:]); print(int(s))"
```

an exception is generated:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/chema/Documents/Workspace/myhdl-git/myhdl/_Signal.py", line 472, in __int__
    return int(self._val)
TypeError: int() argument must be a string or a number, not 'NoneType'

In both cases, the Signal has been initialized to a value, but in the Tristate one, the None becomes the default value.

In the master branch it does not affect, and I am looking for the differences in my copy.

Thanks in any case

PR #80, opened 2015-05-25
ConcatSignal

josyb:
added support for intbv with width of 1
added cast to unsigned if Signal is a port of type std_logic_vector

### DIFF
--- a/myhdl/_ShadowSignal.py
+++ b/myhdl/_ShadowSignal.py
@@ -34,8 +34,8 @@ from myhdl._simulator import _siglist
 from myhdl._bin import bin
 
 # shadow signals
-        
-        
+
+
 class _ShadowSignal(_Signal):
 
     __slots__ = ('_waiter', )
@@ -50,7 +50,7 @@ class _ShadowSignal(_Signal):
         raise AttributeError("ShadowSignals are readonly")
 
 
-        
+
 class _SliceSignal(_ShadowSignal):
 
     __slots__ = ('_sig', '_left', '_right')
@@ -85,7 +85,7 @@ class _SliceSignal(_ShadowSignal):
             yield sig
 
     def _setName(self, hdl):
-        if self._right is None:       
+        if self._right is None:
             if hdl == 'Verilog':
                 self._name = "%s[%s]" % (self._sig._name, self._left)
             else:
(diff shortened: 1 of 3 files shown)


jandecaluwe:
Am I correct in assuming that if the `toVHDL.std_logic_ports` parameter works fine, all the `numeric` checks can go away?

josyb:
Yes, absolutely. My PR kind-of crossed your commit.I can remove this check from the above patch and commit this, or close this PR, clean up my branch and create a new PR?

jandecaluwe:
I saw that you removed some checks already, so I merged this in.

josyb:
There are still a few left in toVHDL.py ... but if we never set _numeric again nothing will break.

jandecaluwe:
I will take care of that as part of a code review. So it's ok for you to remove `toVHDL.numeric_ports`, right?

josyb:
I'll try some more conversions, but I'm sure you can remove ```toVHDL.numeric_ports```. 
I found one issue where I used a ShadowSignal slice of an input std_logic_vector, but I was able to repair that by using the standard slicing method.

josyb:
It would be wise to remove ```toVHDL.numeric_ports``` because if set to false it messes up the ```toVHDL.std_logic_ports``` feature. This may break some code (a lot of mine) but the longer one waits the more it hurts?

josyb:
So far I only found an issue with ShadowSignal, e.g.:
```python
def ST_packntom(Clk, Reset,
                D, DSoP, DEoP, validIn, readyIn ,
                Q, QSoP, QEoP, validOut, readyOut):
...
    if WIDTH_ACCU == WIDTH_D:
        accud = D(WIDTH_D, 0)
    else:
        accud = ConcatSignal(D,accu(WIDTH_ACCU, WIDTH_D))       
...
            if incrfillcount:
                accu.next = accud

```
If *D* is a std_logic_vector port ShadowSignal will use *D* as the name and not the expected *D_num*.
```VHDL
			if bool(incrfillcount) then
				accu <= D(30 - 1 downto 0);
```
Going through the code and my trace.log:  the _setName() in _ShadowSignal.py, that creates *D(30-1 downto 0), is called before the ```toVHDL.std_logic_ports``` are applied. The toVHDL() function doesn't get called at all.

I was able to work around by using an *always_comb* process in stead replacing:
```python
        accud = D(WIDTH_D, 0)
```
with
```python
        accud = Signal(intbv(0)[WIDTH_D:])
        @always_comb
        def assignaccud():
            accud.next = D
```

PR #81, opened 2015-05-25
Conversion py3

jmgc:
Modified the different conversion and test files to make them compatible with PY2 and PY3.

### DIFF
--- a/myhdl/_compat.py
+++ b/myhdl/_compat.py
@@ -1,5 +1,8 @@
+from __future__ import print_function
+from __future__ import division
 import sys
 import types
+from ast import PyCF_ONLY_AST
 
 PY2 = sys.version_info[0] == 2
 PYPY = hasattr(sys, 'pypy_translation_info')
@@ -31,3 +34,7 @@ else:
 
     to_bytes = _identity
     to_str = _identity
+
+def ast_parse(s):
+    return compile(s, '<string>', 'exec', \
+                   print_function.compiler_flag|division.compiler_flag|PyCF_ONLY_AST)
--- a/myhdl/_compat.py
+++ b/myhdl/_compat.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+from __future__ import division
 import sys
 import types
 from ast import PyCF_ONLY_AST
@@ -34,4 +36,5 @@ else:
     to_str = _identity
 
 def ast_parse(s):
-    return compile(s, '<string>', 'exec', PyCF_ONLY_AST)
+    return compile(s, '<string>', 'exec', \
+                   print_function.compiler_flag|division.compiler_flag|PyCF_ONLY_AST)
--- a/myhdl/_compat.py
+++ b/myhdl/_compat.py
@@ -1,5 +1,3 @@
-from __future__ import print_function
-from __future__ import division
 import sys
 import types
 from ast import PyCF_ONLY_AST
@@ -36,5 +34,4 @@ else:
     to_str = _identity
 
 def ast_parse(s):
-    return compile(s, '<string>', 'exec', \
-                   print_function.compiler_flag|division.compiler_flag|PyCF_ONLY_AST)
+    return compile(s, '<string>', 'exec', PyCF_ONLY_AST)
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -52,7 +52,7 @@ from myhdl.conversion._analyze import (_analyzeSigs, _analyzeGens, _analyzeTopFu
 from myhdl._Signal import _Signal,_WaiterList
 from myhdl.conversion._toVHDLPackage import _package
 from myhdl._util import  _flatten
-from myhdl._compat import integer_types, class_types, StringIO
+from myhdl._compat import integer_types, class_types, StringIO, PY2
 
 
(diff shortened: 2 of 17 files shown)


myhdl/_compat.py
```diff
@@ -31,3 +34,7 @@ def to_str(b):
 
     to_bytes = _identity
     to_str = _identity
+
+def ast_parse(s):
+    return compile(s, '<string>', 'exec', \
+                   print_function.compiler_flag|division.compiler_flag|PyCF_ONLY_AST)
```
jck: I think this approach is wrong. This function will always parse the AST with print_function and division enabled even if the particular source file did not import print_function/division from `__future__`

jmgc: Yes, that’s right, but I do not see any major issue with that because the standard print statement will continue working, and it is taken into account in the visit_Print functions on the _toVerilog and _toVHDL files.

In any case, it is quite possible that I missed something.

jmgc:
Dear @jck,

I have checked to remove the flags. You can see the errors generated:

https://travis-ci.org/jandecaluwe/myhdl/jobs/64059365

The problem seems to come from the test_errors and test_ops and test_signed checks. The first and the third due to the print function and the second due to the division. I think that if we want compatibility between both versions of Python, we need these flags.

jandecaluwe:
Wait a moment here - would this type of PR break conversion for existing Py2K style code?

jmgc:
From my point of view no. The only thing, is that the import __future__ for print and division would be needed in those places where they are used. Taking into account that print is used in tests, and division is hardly used, this should not represent a major problem, but I am just one user.

I do not know if @jck has another point of view.

jandecaluwe:
For things like this guessing and opinions don't count. We have to be certain, or at least think we are certain. I am getting very confused about this PR.

jmgc:
Dear @jandecaluwe,

I am not going to tell that I am fully sure that this pull won't imply any risk. I only can give you some facts. I have been using this conversion on a daily basis since myhdl core was made compatible with py3. All the files I have developed have been checked with py2 and py3 and I have not found any problem. All the tests in myhdl work properly, only the prints have had to be modified.

But of course any change implies some burden. In this case, it will imply converting print statements to print functions. I do not mention division as it is seldom used in FPGAs. As I have said, I have not found any other issues.

I think the only solution is that you make some tests with your own code.

jck:
@jmgc,
My point is that applying these compiler flags to all ast parsing is not correct, they should only be applied when necessary.

jandecaluwe:
The proposed patches may have good ideas that we may reuse in the future. However, they clearly are not thought through carefully enough. PRs should be "obvious" improvements that improve things, reduce risk, and do not put the burden of checking things upon others.

jmgc:
Dear @jandecaluwe,

Just a small comment. In any kind of project, to ensure the quality, it is recommended that the tests are done by different people than the one that have made the changes, because everybody makes mistakes, and I am do many a day. This is the reason why I usually ask for a second opinion.

I will not delete the branch in case someone is able to find an alternative for ast_parse.

In any case, thanks for creating this framework.

jandecaluwe:
The way to ensure the quality is working an a feature with those interested in a separate fork, not by raising a premature PR here to get the ball rolling. If I don't accept it, people are disappointed, but if I do and it creates problems, I have an additional maintenance headache. I always lose.

But I am totally swamped. I run 2 (non-MyHDL unfortunately) chip design projects in parallel, and 2 open-source projects. I am therefore looking for contributors that help in *minimizing* my work. And I think I can judge PRs in that respect reasonably well by reading them, also non-trivial ones.

jmgc:
Dear @jandecaluwe,

Here you have the fork: https://github.com/jmgc/myhdl/tree/numeric, it has been tested quite thoroughly, you can see it on travis. I created conversion-py3 just to copy the minimum things needed. It has also been tested, and works.

I understand your concerns, and also @jck ones. I do not agree with them, but these are system decisions that I cannot take. From my point of view the pull shall be the way it is, that's the reason why I have made the pull. If you think they are not, or it must be modified, for me it is right, you are the owner.

In any case, I will continue working on my own fork, and give as much help as possible. The fork numeric will be there for many years, and will be updated and improved, as it is part of a bigger project, but I doubt I will make any other pull to master.

As I have already said, thanks for your efforts.

myhdl/conversion/_toVHDL.py
```diff
@@ -782,6 +782,15 @@ def visit_BoolOp(self, node):
         self.write(")")
 
     def visit_UnaryOp(self, node):
+        if isinstance(node.op, ast.USub):
```
jck: @jmgc This change seems to fix test_numass.py for VHDL.
Could you explain it to me?

myhdl/conversion/_toVHDL.py
```diff
@@ -782,6 +782,15 @@ def visit_BoolOp(self, node):
...
+            if isinstance(node.operand, ast.Num):
+                n = node.operand.n
+                newnode = copy(node.operand)
+                newnode.n = -n
+                newnode.vhd = node.vhd
+                newnode.vhdOri = node.vhdOri
+                self.visit(newnode)
+                return
```
jck: @jmgc This change seems to fix test_numass.py for VHDL.
Could you explain it to me?

Issue #82, opened 2015-05-27
Interface conversion broken in the master branch


cfelton:
I have a [project](https://github.com/cfelton/minnesota) that fails with the latest code on the master branch.  According to git bisect the bug was introduced in 08519b4 [rename attribute ref in the AST before analysis](https://github.com/jandecaluwe/myhdl/commit/08519b452f153885d3cc43038111b7779e2bb2f1).

Warning: this issue simply describes  an observed failure.  Other than using `git bisect` to determine when the bug was introduced, no additional debugging as been completed or a standalone test case created.  Also, the code linked above passes travis-ci because it clones a previous version of the `master` branch (i.e. before 08519b4, it clones 88cb76b).

jmgc:
I detected also this failure (see #64). As far as I know @jck is working on solving it.

Meanwhile, I solved the problem substituting:

```python
        name = node.value.id + '_' + node.attr
        new_name = next(s for s in _suffixer(name) if s not in self.data.symdict)
        self.data.symdict[new_name] = attrobj
        self.data.objlist.append(new_name)
```

by

```python
        new_name = node.value.id + '_Dot_' + node.attr
        if new_name not in self.data.symdict:
            self.data.symdict[new_name] = attrobj
            self.data.objlist.append(new_name)
        else:
            pass
```

Things seem to work well enough, though as I've said, it is a workaround.

cfelton:
@jmgc, thanks for connecting the issues and the suggested workaround.  I don't need any of the upstream changes, using 88cb76b works for me (I just need to be careful between `mn` work and `myhdl` work, probably time to tame virtualenv :).

@jmgc the [gist link](https://gist.github.com/0fb34c528b9ab59a0256.git) to your test example seems to be dead?

jmgc:
Ups, I added a .git after the link. The good one is:

https://gist.github.com/jmgc/0fb34c528b9ab59a0256

Thanks

cfelton:
@jmgc, @jck here is a simplified test the produces the error: https://gist.github.com/cfelton/a6b6457b844c5ab9778f

In the near future I should be able to incorporate this into our test suite.

@jck tangential but relevant, is there an automate method to list the number of tests that are skipped on the myhdl repo landing page?  I realize the `py.test` summary indicates the number test skipped but is there a method for travis-ci to include the number skipped in the myhdl travis-ci banner?

jmgc:
@cfelton, thanks for the effort.

[...]

PR #83, opened 2015-05-28
Added test for issue #82

cfelton:
Created a test for issue #82.  This test reproduces the incorrect multiple driver error.  There is some discussion regarding this error in #64.

Also added an `__init__.py` to test/conversion/ so the `conftest` module can be accessed in the tests.  This is useful when executing the tests individually during development/debug.   

This test currently uses the `bug` decorator, this test will not cause a failure in travis-ci but is flagged as an 'x' test by py.test.

### DIFF
--- a/myhdl/test/conversion/general/test_interfaces3.py
+++ b/myhdl/test/conversion/general/test_interfaces3.py
@@ -11,7 +11,6 @@ from myhdl.conversion import analyze, verify
 
 from myhdl import *
 
-from conftest import bug
 
 class Intf1:
     def __init__(self, x):
--- a/myhdl/test/conversion/general/test_interfaces4.py
+++ b/myhdl/test/conversion/general/test_interfaces4.py
@@ -118,7 +118,7 @@ def c_testbench_one():
         reset.next = not ra
         yield clock.posedge
         for ii in range(10):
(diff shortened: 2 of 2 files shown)


jmgc:
Dear @jck,

I do not want to be a pain, but I have checked your tests with my fork. I think I have detected two "bugs" in the test_interfaces4.py code. They seem to come from the differences between toVerilog and toVHDL conversion. I give you them:

* In the first, the structure:

```python
            assert sdo == expected[ii]
```

My proposal is just change it to:

```python
            t_expected = expected[ii]
            assert sdo == t_expected
```

The second is the structure:

```python
    @always(delay(3))
    def tbclk():
        clock.next = not clock
```

I propose the one from test_pull_64:

```python
        clock.next = 0
        while True:
            yield delay(3)
            clock.next = not clock
```

Finally, I think it is better to initialize the variables inside an instance to avoid unknown states. In this sense, I propose this slight change:

```python
    def tbstim():
        sdi.next = 0
```

Thanks for your patience

cfelton:
@jmgc thanks for reviewing this, yes these are good suggestions.  I verified that the module caused the conversion error but I failed to verify the testbench converted correctly (this is a problem I constituently repeat - valid convertible testbenches, think I would know better by now :)

PR #84, opened 2015-05-29
Std logic ports shadow signals

josyb:
See #80 where I describe an Issue with ShadowSignals from a top-level std_logic_vector port.
As a result of moving the _analyzeTopFunc() forward the original renaming of conflicting generated signals from an interface object is now skipped.
I find silently renaming a less nice solution and added code to raise an error. Also read #64. IMHO I think this will be very rare, and if it arises I prefer changing the naming myself rather than see that (dare I say awkward)  mangled name in the VHDL code.

### DIFF
--- a/myhdl/_ShadowSignal.py
+++ b/myhdl/_ShadowSignal.py
@@ -84,16 +84,20 @@ class _SliceSignal(_ShadowSignal):
             set_next(self, sig[left:right])
             yield sig
 
-    def _setName(self, hdl):
+    def _setName(self, hdl, convertoplevelport):
         if self._right is None:
             if hdl == 'Verilog':
                 self._name = "%s[%s]" % (self._sig._name, self._left)
             else:
-                self._name = "%s(%s)" % (self._sig._name, self._left)
+                self._name = "%s%s(%s)" % (self._sig._name, '_num' if convertoplevelport and self._sig._toplevelport else '', self._left)
         else:
             if hdl == 'Verilog':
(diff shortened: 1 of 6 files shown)


jandecaluwe:
Am I correct in thinking that this PR combines a number of unrelated things (and that the title doesn't reflect everything covered)? In that case, can we please restrict PR's to a single topic only>

As for the std_logic_ports issue - I think the solution is way too complicated and hard too maintain (suppose the rename extension changes). A signal knows about its slice signals - there is an argument for that. So when we rename  a signal like is done for std_logic_ports, we could just add a two-liner loop that sets the names for slice signals again, and done.

josyb:
Actually one thing followed after the other, and perhaps I should have stopped one change earlier.
I too poised about the 'magic' extension name, and how to maintain consistency. 
I overlooked the _slicesignals argument ... it is not clear to me what exactly goes on there (not directly obvious to me from reading the code)
I'll close this PR (I opened it mainly to get a discussion going). And see if I can place that two liner loop?

jandecaluwe:
I think all that should be needed is:

```        
for sl in s._slicesigs:
     sl._setName(hdl)
```

right after the place where a signal's name is modified to support std_logic_ports.

josyb:
I deleted the branch (both local and remote), created a local branch with the same name, added the instructed code, committed and pushed it. But once the branch has been deleted, the PR cannot be re-opened. A new PR is cleaner, anyway?
I checked the function on my MyHDL modules, should there be a formal test?

jandecaluwe:
A new PR is cleaner.

PR #85, opened 2015-05-30
fixed convertible testbench issues

cfelton:
The new test (test_interfaces4.py) correctly verified the module converted but had errors in the convertible testbench, the testbench was functionally correct but had non-convertible constructs.  Those issues are fixed in this pull-request.


jandecaluwe:
I'm not fond of adding tests without the fix to the official repo. If the test has problems, it creates more problems than it solves for whoever want to implement the fix.

cfelton:
@jandecaluwe fair enough, will take note for the future.

jck:
I like such tests, it makes it simpler for me to work on the fix.

jandecaluwe:
Fine!

PR #88, opened 2015-06-23
added key in dict check before access

cfelton:
Was reviewing some code and ran into a complicated modeling case when a signal was buried deep inside a combination of dictionary and classes.  The node being evaluate was not in the symbol dict.  Adding the key check before accessing resolved the issue.


jck:
Chris,
Can you try that case with my issue_82 branch?
( `pip install https://github.com/jck/myhdl/archive/issue_82.zip` in a virtualenv)

cfelton:
@jck same error with the issue_82.zip version tested in a virtualenv.

jandecaluwe:
I am worried that this change would mask real issues?

cfelton:
@jandecaluwe would adding a warning be sufficient?  My thought is this is a special case where Signals, used in modeling, are embedded in complex structures that are not convertible.  The chances of seeing this with convertible interfaces is unlikely.

I agree, I don't know the mechanism why certain signals are add to the dict and some are not (the worry is warranted, the actual fix would verify the absence of a signal is expected).

jck:
@cfelton Could you provide an example which triggers this issue? I will try fixing it.

cfelton:
Closing this pull-request per the comments, need a more general solution.

PR #89, opened 2015-06-26
Windows support for co-simulation

### LINKED ISSUE #38: Cosimulation error with python 3.2
MyHDL does not seem to like running with Python 3 earlier than 3.4:

            # New pipes are not inheritable by default since py 3.4
            if not PY2:
                for p in rt, wt, rf, wf:
    >               os.set_inheritable(p, True)
    E               AttributeError: 'module' object has no attribute 'set_inheritable'
    
    /usr/local/lib/python3.2/dist-packages/myhdl/_Cosimulation.py:61: AttributeError

iscgar:
Added support for windows pipes to allow co-simulation in the Windows version of ModelSim.
Currently only the ModelSim bridge has been updated because this is what I needed to work with. Initial tests show that it's still building and working as expected on Linux, but more test environments are needed to make sure nothing is broken because of this.

Also fixes #38

### DIFF
--- a/cosimulation/modelsim-win/Makefile
+++ b/cosimulation/modelsim-win/Makefile
@@ -1,0 +1,21 @@
+INCS := C:\\modeltech64_10.4\\include
+LIB_PATH := C:\\modeltech64_10.4\\win64
+
+ifneq ($(filter cl%,$(CC)),)
+	CFLAGS := /LD /I$(INCS)
+	LIBFLAGS := $(LIB_PATH)\\mtipli.lib
+else
+	CFLAGS := -static -g -I$(INCS) -o myhdl_vpi.dll
+	LIBFLAGS := -L$(LIB_PATH) -lmtipli
+endif
+
+all: myhdl_vpi.dll
+
+myhdl_vpi.dll: myhdl_vpi.c
+	$(CC) $(CFLAGS) $< $(LIBFLAGS)
+
+clean:
+	@del /q myhdl_vpi.dll || rm -rf myhdl_vpi.dll 2>&1> /dev/null ||:
+	@del /q myhdl_vpi.lib || rm -rf myhdl_vpi.lib 2>&1> /dev/null ||:
+	@del /q myhdl_vpi.exp || rm -rf myhdl_vpi.exp 2>&1> /dev/null ||:
+	@del /q myhdl_vpi.obj || rm -rf myhdl_vpi.obj 2>&1> /dev/null ||:
--- a/cosimulation/modelsim-win/myhdl_vpi.c
+++ b/cosimulation/modelsim-win/myhdl_vpi.c
@@ -1,9 +1,9 @@
 
 #include <stdlib.h>
-#include <unistd.h>
 #include <assert.h>
 #include <string.h>
 #include <stdio.h>
+#include <Windows.h>
 #include "vpi_user.h"
 #include "sv_vpi_user.h"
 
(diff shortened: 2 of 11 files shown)


jandecaluwe:
I don't like the fact that all kind of unrelated complications sneak in to support Python < 3.4 while it was decided not to do so. I also don't like that the modelsim vpi for for linux is modified, with a corresponding maintenance risk. Given the amount of non-standard issues, both in Verilog vpi and in Linux vs Windows, it is much better to consider this to be a separate simulator with its own dedicated vpi file.

iscgar:
I removed the < 3.4 support code, but I had to add a `hasattr` check to trick Travis into accepting the build. We either need support for < 3.4 or these scenarios should be removed.
Also moved the ModelSim code for Windows into its own directory.

cfelton:
@iscgar when you say "these scenarios should be removed" what are you referring to?

travis-ci is setup to test the following python versions:
```
python:
  - "2.6"
  - "2.7"
  - "pypy"
  - "3.4"
```

Regards,
Chris

iscgar:
@cfelton I mean that if the required python version is >= 3.4 as @jandecaluwe said, then there's no point in testing against 2.6, 2.7 and PyPy (not even PyPy3 which only implements 3.2.5).

Though I'd much rather have support for < 3.4 since os.set_inheritable seems to be the only 3.x feature used and it'd be shame to drop support when we can simply emulate it.

jandecaluwe:
I **obviously** meant that Python3K will not be supported for versions < 3.4.

cfelton:
@iscgar as Jan stated, python 2 support is needed (not going away) but the recent(ish) python 3 support is limited to >= 3.4 (similar python 2 support is >= 2.6).  Yes, agreed the pypy3 won't be officially supported until it reaches 3.4.

I don't know the details of limiting support >= 3.4 (or I forget) but it is broader than this topic/change.

iscgar:
@jandecaluwe Well, in that case, without having some way to make `HANDLE`s inheritable there will never be Windows support for python < 3.4 unless `Popen` is replaced with one of the `spawn*` functions -- in which case closing the unneeded pipe ends would have to be done from the C code because of the `fd`s will be inherited..

jck:
@iscgar os.set_inheritable is not the only 3.x feature which caused us to make the decision. 
There were quite a few AST related changes which were introduced in 3.4. see http://greentreesnakes.readthedocs.org/en/latest/nodes.html

jandecaluwe:
Travis test fails.

Also, _Cosimulation.py has a number of potentially subtle changes. Why is the check on PY2 now no longer necessary? Why is  the set_inheritable handling now so different? Why is the preexec_fn call dropped??? These are recent changes that passed all tests - I must assume they we introduced for very good reasons.

iscgar:
@jandecaluwe I removed the PY2 check because I thought that you meant that only >= 3.4 is supported, so checking for 2.x didn't make sense for me, but used `hasattr` for the Travis checks.

I'm not sure what's causing the failure with the icarus tests, but since it was passing with the set_inheritable code that I set in place in the previous commit, so I can only assume that maybe the unneeded pipe ends that are left open in < 3.4 somehow interfere with the communication.

Now, if python 2.x is still supported, all the supporting code I removed should be restored because pipes created using `os.pipe` were always non-inheritable on Windows so there needs to be an explicit action to make them inheritable even in < 3.4. The `set_inheritable` handling was changed because of that and the `preexec_fn` call was dropped because it was no longer necessary since the unneeded pipe ends weren't inherited by the child process.

Should I refactor the set_inheritable code and put it in the _util module to make it work? Otherwise, as I stated earlier, Windows support wouldn't be possible it we stick to pipes.

iscgar:
@jandecaluwe Travis tests are now passing. I cleaned-up the code in _Cosimulation.py and extracted all the gory details that are needed to support Windows into the _util module.

jandecaluwe:
@jck, what do you think about the changes to _Cosimulation.py?

jck:
I'd like to see some links/references in the comments of the low level set_inheritable implementations.
Apart from that, it looks good.

iscgar:
@jck added references to the code and proposal on which the set_inheritable code is based on.

Issue #96, opened 2015-07-09
Using parameters for delays doesn't work with simulator vsim


d-roma:
I'm new at myhdl so most probably I'm doing something wrong but from the rs232_tx example I think what I'm trying should be possible. 

I have defined a constant with an integer value. Then, I use this constant for the delay function. 

```python
HALF_PERIOD = 5

@instance
def clockGen():
    yield delay(HALF_PERIOD)
    ...
```

Everything seems to work fine with the testbench in python, but when I try to test the vhdl, using the vcom simulator (documentation says vsim but in mine _verify module it's defined as vcom) I get the following error:

> ** Error: testResetValues.vhd(53): near "ns": expecting ';' 

The reasons is that the resulting vhdl has the following code:

```vhdl
constant HALF_PERIOD: integer := 5;
...
    wait for HALF_PERIOD ns;
```
As far as I known this is not supported for vhdl. Either the constant should be of type time instead of integer and the time unit inserted at the definition or the value should be directly replaced after the wait for.

cfelton:
@d-roma Thanks for reporting this, I think this is a known issue but hasn't been recorded anywhere. 

Here is a past mailing-list thread that discusses a fix: http://thread.gmane.org/gmane.comp.python.myhdl/3863/focus=3865

d-roma:
You're righ! As reported by http://thread.gmane.org/gmane.comp.python.myhdl/3863/focus=3865
changing line 1151 of _toVHDL.py from 

```python
            self.write(" ns")
```

to 

```python
            self.write("*1.0 ns")
```

Solved the issue for me. Thanks! Perhaps it would be good to commit this change, I don't can see any case where this could get in conflict with anything else.

jandecaluwe:
Is there a reason why it has to be 1.0 ns instead of just 1 ns?

d-roma:
No, no one, it also works fine with just 1 ns. Even if at some point the multiplying constant changes from integer type to real, it will still work using 1 ns. So, for vcom it seems completely indifferent using just 1 or 1.0. But probably using just 1 ns will ensure being compatible with every simulator, since the VHDL standard always uses integer values for physical types.

cfelton:
@jandecaluwe, @d-roma I was working off @josyb suggestion in the original mailing-list thread.  We want to use "1' instead of "1.0" based on the integer is the common value for VHDL physical types?

cfelton:
Updated the PR #97 to use "1".  

[...]

PR #97, opened 2015-07-09
VHDL delay (wait) conversion fix

cfelton:
This fixes the issue when a constant is used for the delay value.  This implements the fix suggested in #96 and the mailing-list thread listed in #96.

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -1006,7 +1006,7 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
             self.write(f.__name__)
         elif f is delay:
             self.visit(node.args[0])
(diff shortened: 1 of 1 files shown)


jmgc:
Dear @cfelton,

Just a small suggestion. I think it would be better to add parenthesis at the beginning and the end. This will make the product operation more "robust".

cfelton:
@jmgc, beginning and end of what?  the string ` * 1.0 ns` will always be the suffix and if necessary the converter will put parentheses around the leading terms, example:

```
yield delay(HALF_PERIOD*5 + 13)
```
is converted to
```
wait for ((HALF_PERIOD * 5) + 13) * 1.0 ns;
```

Regards,
Chris

jmgc:
Then, nothing more to say.

Thanks for the quick answer

Issue #98, opened 2015-07-09
TristateSignal Conversion Error and Verilog missing inout


cfelton:
When `TristateSignal`s are used as top-level ports the signal is not converted to an `inout` when the target is Verilog.  This appears to work in VHDL but not in Verilog. 

In addition PR #74 appeared to break [code that use to convert](https://gist.github.com/cfelton/6119313).  I used `bisect` to verify the revision where the code stopped converting f965132.

Failing code snip:
```python
def top(sda, scl, sda_i, sda_o, scl_i, scl_o):
     sda_d, scl_d = sda.driver(), scl.driver()
     @always_comb
     def hdl():
         sda_i.next = sda
         sda_d.next = False if not sda_o else None 
         scl_i.next = scl
         scl_d.next = False if not scl_o else None
     return hdl
```

It is not clear (at this point) why the conversion is failing.  The conversion for the code snip only fails for VHDL conversion.

josyb:
It will convert (for VHDL) if you change the ternary construct:
 ```python
       sda_d.next = False if not sda_o else None
```
into:
```python
        if sda_o:
            sda_d.next = None
        else:
            sda_d.next = False
```

josyb:
It looks like the AST traversal errs:
This construct breaks at encountering the 'None' 
```python
sda_d.next = False if not sda_o else None
```
```text
line: sda_d <= '0' when (not bool(sda_o)) else stdl( 
```
and if we swap places:
```python
sda_d.next = None if sda_o else 0
```
```text
line: sda_d <= stdl(
```
Note that here the 'stdl(' is wrong too, as the expected first object in  the eventual ' a when b else c' statement is a 'Z'.
#74 corrected (tried to) the invalid (others => 'Z') when the RHS had a width of 1 (bool or intbv[1:]). If we have an ternary construct, this width information is missing.

jmgc:
Dear @josyb,

I think this will be useful inside procedures, but not in the concurrent part. That's the reason why VHDL2008 would be the quickest solution, but it is not available right now.

Best

josyb:
@jmgc, dear Jose,

actually the ' a <= b when c else d ;' is a valid concurrent statement in VHDL ( a conditional signal assignment statement). MyHDL will convert the original code into a series of concurrent statements.
The issue here is that MyHDL has a problem converting the python 'b if c else d' construct.

Regards,

Josy

jmgc:
[...]

PR #100, opened 2015-07-10
Proposal to solve the issue #98 on VHDL

jmgc:
The proposed solution is to have also a check for a None object in infercast for the ternary operator, and add a check for None in the visit_NameConstant to avoid errors. A test has also been added in test_ternary.


cfelton:
@jmgc the travis-ci build seems to be unrelated to the code, is seems that it failed to install GHDL (and oddly during an icarus test).  

Also, the changes in this PR are minimal but the change log is really long?  Is this the result of an odd git merge or something?  

It looks like the solution and test is reasonable and this would be a good fix before the 0.9.0 release?

jmgc:
Yep, you are right, I have made quite an odd pull request, but I do not know how to avoid this "weird log". Also the Icarus failure only happens in version 2.7. Any ideas on how to get out of this mess?

Thanks in advance

cfelton:
@jmgc I think (looked quickly) the failure was independent of the python version and code, it failed while trying to retrieve GHDL from an external server (the request hung?).  

I don't know how to clean up the git log?  If you make a small change on the git branch it will force travis-ci to redo the tests.

jmgc:
I have made a small change (remove a line return) and now it seems to work in travis.

cfelton:
@jmgc we will need to fix this, we are relying on travis-ci to be accurate, it is our gate for PRs.  If it fails for installing packages (independent of the changes) it won't serve us too well.  Need to add it to the list (at least it doesn't appear to happen too frequently).

jandecaluwe:
Do you really expect me to consider a PR with 72 unrelated commits?

jmgc:
No, of course, I was just waiting to know your opinion about the changes. Depending on that, I will try to find a solution for the log, because something tells me it will take me some time to solve it.

Best,

Jose M.

jmgc:
Dear Jan,

Please, reject the PR (I have no access to my computer). Apart from the log issue, there is a case that @josyb has pointed out and I have fully forgotten intbv[1:].

Sorry for the inconveniences,

Jose M.

PR #101, opened 2015-07-12
[RFC] Proposal to solve the issue #98 on VHDL, substitutes PR #100

jmgc:
Dear all,

Based on the comments on issue #98 and PR #100, I have create a new PR. It includes tests on both languages. Please @cfelton and @josyb, if possible, confirm that it works with your files.

Best

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -1124,17 +1124,12 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
             self.write(';')
 
     def visit_IfExp(self, node):
-        pre, suf = self.inferCast(node.vhd, node.body.vhdOri)
-        self.write(pre)
+        node.body.vhd = node.orelse.vhd = node.vhd
         self.visit(node.body)
-        self.write(suf)
         self.write(' when ')
         self.visit(node.test)
         self.write(' else ')
-        pre, suf = self.inferCast(node.vhd, node.orelse.vhdOri)
-        self.write(pre)
         self.visit(node.orelse)
-        self.write(suf)
 
     def visit_For(self, node):
         self.labelStack.append(node.breakLabel)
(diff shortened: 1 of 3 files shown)


cfelton:
@jmgc I will review and test this against my code snippets, eventually.

My time will be serverly limited the next week.  Also, given the feedback in #102 and the mailing-list you should add [WIP] or [RFC] to the title/subject of this PR.

Also, I would like to inlcude the Verilog fixes in this PR, when I have time I call pull from your branch, commit my changes to the branch, and then you can pull from my branch and the PR will be updated with the changes.

jmgc:
Changed to [RFC]

jandecaluwe:
I think this is overly complicated. I don't like the checks to `None` and the check/manipulation of a specific convertor. Most importantly, I have commented on #98 before and pushed a commit which I think solves the issues in a better way.

cfelton:
@jandecaluwe sorry I missed the #98 comment and [commit](https://github.com/jandecaluwe/myhdl/commit/14dc28f79870a5bade8d5f12a9f70db61f3ac083), could have saved some time and effort.  I tested 0.9.0 with my examples and there are no conversion errors (the conversion looks reasonable), thanks.

jmgc:
Dear Jan,

I answer inlined.

> El 15/07/2015, a las 18:15, jandecaluwe <notifications@github.com> escribió:
> 
> I think this is overly complicated.
> 
I am sure it would not have taken so much time an comments if it was so simple. Anyhow, the changes are minimal, and try to reuse code a much as possible.
> I don't like the checks to None and the check/manipulation of a specific convertor.
> 
Me neither, but the high impedance state has been defined using None, and there is not much room for changes. Or could be yes...
> Most importantly, I have commented on #98 before and pushed a commit which I think solves the issues in a better way.
> 
Yes, it is true that you provided an alternative, but you did not do you homework, assumed that it was trivial, and did not provide a test to check it. If you had done it, you would have seen that some other changes were needed. But, as I am disciplined, I did it and modified the code to solve the issues.

Now you have the tests in travis that failed previously, and work now.

Best

PR #102, opened 2015-07-13
Initial value support

hgomersall:
Following cfelton's initial efforts and some discussion, this patch is adds initial value support to the toVHDL and toVerilog conversions.

Included are tests against Xilinx Vivado, which are all satisfied - these are necessarily non-vpi in nature (which is not supported by Vivado), so I've used the testing framework I developed for working with Vivado (veriutils). If Veriutils is not present, it gracefully fails and skips the tests.

Initial value support is enabled by setting disable_initial_value = False (which is default True). I suggest that the default being True allows a soft entry to asserting compatibility with the various toolchains as people have the time and inclination to test them.

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -100,6 +100,7 @@ class _ToVHDLConvertor(object):
                  "use_clauses",
                  "architecture",
                  "std_logic_ports",
+                 "disable_initial_value"
                  )
 
     def __init__(self):
@@ -113,6 +114,7 @@ class _ToVHDLConvertor(object):
         self.use_clauses = None
         self.architecture = "MyHDL"
         self.std_logic_ports = False
+        self.disable_initial_value = True
 
     def __call__(self, func, *args, **kwargs):
         global _converting
@@ -403,8 +405,10 @@ def _writeSigDecls(f, intf, siglist, memlist):
                               category=ToVHDLWarning
                               )
             # the following line implements initial value assignments
-            # print >> f, "%s %s%s = %s;" % (s._driven, r, s._name, int(s._val))
-            print("signal %s: %s%s;" % (s._name, p, r), file=f)
+            if toVHDL.disable_initial_value:
+                print("signal %s: %s%s;" % (s._name, p, r), file=f)
+            else:
+                print("signal %s: %s%s = %s;" % (s._name, p, r, int(s._val)), file=f)
         elif s._read:
             # the original exception
             # raise ToVHDLError(_error.UndrivenSignal, s._name)
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -405,10 +405,22 @@ def _writeSigDecls(f, intf, siglist, memlist):
                               category=ToVHDLWarning
                               )
             # the following line implements initial value assignments
+
+            sig_vhdl_obj = inferVhdlObj(s)
+
             if toVHDL.disable_initial_value:
-                print("signal %s: %s%s;" % (s._name, p, r), file=f)
+                val_str = ""
             else:
-                print("signal %s: %s%s = %s;" % (s._name, p, r, int(s._val)), file=f)
+                
+                if isinstance(sig_vhdl_obj, vhd_std_logic):
+                    val_str = " := '%s'" % int(s._init)
+                elif isinstance(sig_vhdl_obj, vhd_int):
+                    val_str = " := %s" % s._init
+                else:
+                    val_str = ' := "%s"' % bin(s._init, sig_vhdl_obj.size)
+            
+            print("signal %s: %s%s%s;" % (s._name, p, r, val_str), file=f)
+
         elif s._read:
             # the original exception
             # raise ToVHDLError(_error.UndrivenSignal, s._name)
--- a/myhdl/conversion/_toVerilog.py
+++ b/myhdl/conversion/_toVerilog.py
@@ -297,7 +297,7 @@ def _writeSigDecls(f, intf, siglist, memlist):
                 print("%s %s%s%s;" % (k, p, r, s._name), file=f)
             else:
                 print("%s %s%s%s = %s;" % 
-                      (k, p, r, s._name, _intRepr(s._val)), file=f)
+                      (k, p, r, s._name, _intRepr(s._init)), file=f)
         elif s._read:
             # the original exception
             # raise ToVerilogError(_error.UndrivenSignal, s._name)
--- a/myhdl/conversion/_toVerilog.py
+++ b/myhdl/conversion/_toVerilog.py
@@ -95,7 +95,8 @@ class _ToVerilogConvertor(object):
                  "no_myhdl_header",
                  "no_testbench",
                  "portmap",
-                 "trace"
+                 "trace",
+                 "disable_initial_value"
                  )
 
     def __init__(self):
@@ -109,6 +110,7 @@ class _ToVerilogConvertor(object):
         self.no_myhdl_header = False
         self.no_testbench = False
         self.trace = False
+        self.disable_initial_value = True
 
     def __call__(self, func, *args, **kwargs):
         global _converting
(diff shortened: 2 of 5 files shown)


cfelton:
@hgomersall when you say Vivado support do you mean Vivado synthesis or Vivado simulation?  If it is the later no external tools are needed for non-vpi testing.  All the tests in `conversion/general` are non-vpi tests.

But with that said, we should have VPI specific tests for initial value support because it was a source of failure (where the MyHDL simulation didn't match the V* simulation).

The other issue was synthesis support.  We need to test the various synthesis tools and verify that the initial value is supported (at one time it wasn't).  Think we need to check off all the following are supported:

    - [ ] Vivado synthesis
    - [ ] ISE synthesis
    - [ ] Quartus synthesis
    - [ ] Lattice Diamond (synplicity) synthesis
    - [ ] Icarus Verilog

Since we are close to a 0.9.0 release this might need to wait until 1.0.

hgomersall:
I agree re more tests and more tools, I just wanted to plant a flag.

I mean Vivado simulation at the moment, but it's not clear to me how the existing tools allow Vivado simulation - specifically, the point is to check that Vivado returns the correct output given a value set from an initial value, hence the external tool. Perhaps I'm missing something here?

jandecaluwe:
The existing test infrastructure should be sufiificent to implement this. I don't see why we need dependencies on Vivado and external tools and I also don't think that is a good idea.

@cfelton Can you perhaps take the lead in this with your experience? I think we will want to enable it by default in the next version (as the converted code then matches the MyHDL behavior better) and use a positive setting, e.g. use_initial_values.

hgomersall:
@jandecaluwe there is no dependency, just the facility to test it (it will happily skip the test if it can't be run). It is useful to have the check that Vivado is doing what is expected. Clearly, there should be other tests for other toolchains, but I see no reason why we shouldn't include the test (I see the potential for verification against other systems as a positive feature).

jandecaluwe:
Is this PR verified by tests running on the travis infrastructure or not?

hgomersall:
Only insomuch as it doesn't break things. It's reasonable to assert that e.g. Icarus and GHDL should support the changes, but in practice the whole point is to target more mainstream tools, especially given one big use of this is in FPGA initialisation. I do think that Icarus and GHDL support should be added before the PR is merged, but I don't think that other verifications should be removed (even if currently Travis can't support them).

jandecaluwe:
I do not agree - at all.

For any new clearly defined feature except trivial ones, there should be new tests that fail before and pass after the feature implementation. And these test should make use of the existing infrastructure. That should be more than enough - the fact that they are "minimal" non-mainstream tools is an advantage. I simply don't see the need to add additional dedicated code that is not part of the infrastructure - all that does is to create an unverifiable maintenance headache.

Finally, I simply do not see it as my job to keep track of the mergeability status of PR. If presented here, it should be ready to be accepted readily or not. You can work in other repos to cooperate and keep things current.

hgomersall:
The problem Jan is that you're very difficult to engage in discussion yet you are the arbiter of what should be accepted. A PR seems to be the only way to actually reliably get your opinion on things. I appreciate you're busy, but there are many people that are keen to contribute and currently we're finding it very difficult.

Of course I wrote the code through tests prior to implementation, which was why I am confident of it's correctness. Now, clearly we disagree as to what should be merged into the code-base, but I'm not precious about this so I'm happy to strip out non-travis compatible tests.

It is unnecessary to close the PR as modifications happen in situ  and the PR is updated accordingly. It's perfectly reasonable that the PR serves as the forum for discussion and modification, converging on an acceptable solution.

jck:
I agree that it is reasonable to use PRs for discussion. For example, the way neovim does it is really good.

---------
When submitting pull requests (commonly referred to as "PRs"), include one of the following tags prepended to the title:

    [WIP] - Work In Progress: the PR will change, so while there is no immediate need for review, the submitter still might appreciate it.
    [RFC] - Request For Comment: the PR needs reviewing and/or comments.
    [RDY] - Ready: the PR has been reviewed by at least one other person and has no outstanding issues.

During the RFC/WIP stage, people can keep updating their PRs and after it has been reviewed, it is tagged as RDY and a colloboarator can merge it in.
---------

In this particular case, when the purpose of the changes is to primarily the benefit of proprietary tools, we need to take peoples word.

Additionally, if using @hgomersall 's library makes it considerably simpler to test Vivaldo support, it is not unreasonable to expect users of Vivaldo to install that if they want to run the particular tests.

jandecaluwe:
The purpose of initial value support is not the benefit of proprietary tools. It is in the first place to have a converted model that matches the MyHDL behavior better. There should be no urgent reason to add additional simulation tool dependencies, it's Verilog/VHDL like everything else. The only reason why initial value support is currently not in there is because of prior limitations to proprietary *synthesis* tools - something we cannot reasonably expect to support with travis.

PR's can be fine to discuss about concrete code and implementation, as I have been doing regularly, and if you want to add staging conventions, fine. But for initial discussions and directions, it is much better to use Issues - and it is demonstratably false that I don't give feedback there if I have something meaningful to say. 

http://dev.myhdl.org/tasks/initial-values-support.html

hgomersall:
I've sent an email to the list relating to github protocols, taking it off this PR.

I'm happy to rework this PR for full Travis support, removing the Vivado tests, but it might take a few weeks.

Just to answer one point about the worth of initial values. Xilinx guarantees and recommends initial values for setting the initial state of the FPGA, at least in the latest generation. Now this is just one vendor, but there is a clear use case which goes well beyond simply making the simulation work as expected, and into actual design considerations. Supporting such implementations is what motivated me, and is why Vivado verification was the first set of tests I implemented.

jck:
Quartus(Altera) has also supported initial values for atleast the last 2 years.

jandecaluwe:
@hgomersall Excellent, I look forward to that PR. Take the time necessary; as we aim for world domination there is no urgency :-) Please consider enabling this by default and use a positive configuration parameter, e.g. `use_initial_values`.

@hgomersall @jck For the record, I am the biggest fan of initial values, regardless of what vendors say about this (which may vary over time) - so much that this was supported in the earliest Verilog conversion code (2005?) because the better behavioral match that avoids certain simulation issues. I reluctantly took this out when a Quartus synthesis user complained.

cfelton:
Couple comments, the `disable_init_values` I proposed (poorly?) to match the current `toV*` attributes.  That being, an attribute that sets the non-default, e.g. `no_myhdl_header`, `no_myhdl_package` and `no_testbench`.  The thought, eventually initial values are the default and only special cases would one disable (the near-term is the opposite).  To be consistent `no_initial_values` should have been proposed.

I don't have a strong feeling either way, but I wanted to provide the background info (above).  I am ok with using `use_initial_values`.  As stated in this thread, the goal will be a default value `use_initial_values = True` for 1.0.

I agree with both points of view regarding the PRs (master waffler :).  That is, the PRs can be a great collaboration tool (inline comments in the code!) but we need processes and methods for efficient communication.  I think the proposed solution could work.  I imagine *staleness* will persist - simply a function of limited resources and distributed development.  An OP needs to be patient and persistent with an PR/issue they feel warrants attention.

For issues, my thought is the author uses the github labels in the title: [bug], [enhancement], and [question].  This way the author can suggest the topic for the issue but the official assignment is deferred.

jandecaluwe:
Can't we use labels? Then it is easy to fitter.

jandecaluwe:
@cfelton You are correct, I was not consistent in my recommendations. `no_initial_values` is indeed a better choice.

hgomersall:
@jandecaluwe I think only you have permission to set labels. If we moved the main repository to an organisation, then there is better flexibility to give others the permission to set labels, but it would still be a restricted set I think. One potential sticking point is that the rights to set labels overlap fully with the rights to accept pull requests: https://help.github.com/articles/permission-levels-for-an-organization-repository/

jandecaluwe:
@hgomersall Right. I didn't realize labels were restricted like that.

jck:
We can easily do a bot which responds to github webhooks, but it will have to be given push access.

hgomersall:
@jandecaluwe Has the initial value stuff been sorted now? I was just revisiting this and it seems to be at least somewhat handled.

hgomersall:
@jandecaluwe actually, scratch my last comment - I misinterpreted a constant driver. I'll get onto this to complete the PR.

PR #111, opened 2015-07-18
Various HDL simulator,  conversion test improvements

jck:
Details in commit messages.


cfelton:
+1 I like the renaming of the simulators to the executable name, I like that we have standardized on `py.test`, and `py.test --sim <simulator>` is a nice feature.

jandecaluwe:
Nice, thanks.

PR #112, opened 2015-07-19
Added Code to remove inappropriate warnings on Tristate Ports and TristateDrivers 

josyb:
The code:
```python
def top( sl, slv, sl_i, sl_o, slv_i, slv_en, slv_o ):

    sl_d = sl.driver()
    slv_d = slv.driver()

    @always_comb
    def hdl():
        sl_i.next = sl
        sl_d.next = False if not sl_o else None
           
        slv_i.next = slv
        slv_d.next = slv_o if slv_en else None

    return hdl
```
when converted, both VHDL and Verilog, will issue a warning for:
   1. the Output Ports  _sl_ and _slv_ being read
   2. the TristateDrivers _sl_d_ and _slv_d_ being driven but not read.

The proposed code adds checks to suppress those warnings.

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -325,10 +325,11 @@ def _writeModuleHeader(f, intf, needPck, lib, arch, useClauses, doc, stdLogicPor
             if convertPort:
                 pt = "std_logic_vector"
             if s._driven:
-                if s._read:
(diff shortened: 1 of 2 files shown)


jandecaluwe:
It seems the intention to check on a signal type, and this is done indirectly by checking the presence of attributes. Such checks should be done directly on the signal subtype itself, using `isinstance()`.

josyb:
Old C-habits die hard :)

PR #113, opened 2015-07-22
core decorator improvements

jck:
* More code reuse across instance, always, always_comb, always_seq
* Groundwork for streamlined AST processing. Each generator's AST is stored as an attribute of its instantiator.


jck:
I've removed the 'remove unused methods' commit

jandecaluwe:
Nice work! thanks.

PR #129, opened 2015-10-02
always_seq, always_comb improvements

jck:


jandecaluwe:
Fascinating stuff! This is obviously a great simplication of the code. So... where did you find out about `cell_contents`? I remember struggling with this in the early days and I got help from a Python guru (Samuele Pedroni) to solve the problem with a hack. Did we overlook `cell_contents`? Or was this introduced in some later Python version (or backported from Python3?)? I searched through the documentation and I haven't yet found much about this attribute.

jck:
I found out about it here: http://stackoverflow.com/questions/26094700/cell-contents-in-python-closure

Python 3.3 introduced inspect.getclosurevars which uses cell_contents under the hood.
https://github.com/python/cpython/blob/d15ca79a9284ab5e1d8fb552e31955a1092f339c/Lib/inspect.py#L1293

PR #133, opened 2015-10-29
assign statement (constwires) support litterals larger than 32 bits

clade:
The size of literal integer was not specified in the assign statement.

### DIFF
--- a/myhdl/conversion/_toVerilog.py
+++ b/myhdl/conversion/_toVerilog.py
@@ -324,7 +324,9 @@ def _writeSigDecls(f, intf, siglist, memlist):
             c = int(s.val)
         else:
             raise ToVerilogError("Unexpected type for constant signal", s._name)
-        print("assign %s = %s;" % (s._name, c), file=f)
+        c_len = s._nrbits
+        c_str = "%s"%c
+        print("assign %s = %s'd%s;" % (s._name, c_len,  c_str), file=f)
     print(file=f)
     # shadow signal assignments
     for s in siglist:


jandecaluwe:
Do you have a testcase that exposes the problem?

clade:
Here is a test case 

```
from myhdl import *

i = Signal(intbv(0, min=0, max=256))
out = Signal(False)
clk = Signal(False)

def test(clk, i, out):
    large_signal = Signal(intbv(123456789123456, min=0, max=2**256))
    @always(clk.posedge)
    def logic():
        out.next = large_signal[i]

    return logic

toVerilog(test, clk, i, out)
```

In the conversion, the large_signal (which is not driven) is converted to a wire an then there is an assign 

```
assign large_signal = 123456789123456;
```

It should be replaced by : 
```
assign large_signal = 256'd123456789123456;
```

jandecaluwe:
I cannot reproduce the problem. With Icarus iverilog, the large integer constant works fine unsized.

clade:
Vivado Design Suite generates the following error :

```
ERROR: [Synth 8-1872] decimal constant 123456789123456 is too large, 
using -2045800064 instead
```

It works well with the 256'd prefix.

I am not an expert in Verilog, but apparently literals with no prefix 
are assumed to be 32bits (see for example 
http://web.engr.oregonstate.edu/~traylor/ece474/lecture_verilog/beamer/verilog_number_literals.pdf).

jandecaluwe:
Thank you.

PR #135, opened 2015-11-08
add test for issue 134

nturley:
checks for port name collision when a function names two signals with the same name


cfelton:
@nturley thanks for adding the test, this can be very helpful.

PR #136, opened 2015-11-13
created a Factory class with static methods for cleaner generation of…

fwswdev:
created a Factory class with static methods for cleaner generation of Signals
(in progress)


nicmcd:
@fwswdev can you say what is not clean already?

jck:
I think we can close this issue.

jck:
@fwswdev You can reopen this issue if you can make a case for why we need this.

PR #144, opened 2016-01-25
Add 'directory' attribute to traceSignals.

nicmcd:
This change does exactly what the 'directory' attribute for toVerilog and toVHDL
do. It allows the user to specify the output directory for VCD files.

This commit includes the source change, documentation change, and a py.test
change to test the functionality.


cfelton:
@nicmcd thanks for a complete PR (docs and all)!  

I realize you used the existing `toV*.directory` attribute as an example.  But the code could probably be simplified using the default assignment as `self.directory=''`.  Not sure why `None` was used in the toV* but I don't think a default of `None` is needed, a valid directory string is sufficient.

PR #145, opened 2016-01-30
Add 'filename' attribute to traceSignals

punkkeks:
Additional to the directory attribute, I'd like to have a filename attribute. The reason is, when I run tests, the resulting vcd files are all named the same, that will result in the unreadable date suffix. 
The reason for not changing the name attribute is, that myhdl uses it to extract the hierachy and will rename it in the vcd file, which is ugly.

### DIFF
--- a/myhdl/_traceSignals.py
+++ b/myhdl/_traceSignals.py
@@ -50,6 +50,7 @@ class _TraceSignalsClass(object):
 
     __slot__ = ("name",
                 "directory",
+                "filename",
                 "timescale",
                 "tracelists"
                 )
@@ -57,6 +58,7 @@ class _TraceSignalsClass(object):
     def __init__(self):
         self.name = None
         self.directory = None
+        self.filename = None
         self.timescale = "1ns"
         self.tracelists = True
 
@@ -89,8 +91,13 @@ class _TraceSignalsClass(object):
             else:
                 directory = self.directory
 
+            if self.filename is None:
+                filename = self.name
+            else
+                filename = self.filename
+
             h = _HierExtr(name, dut, *args, **kwargs)
-            vcdpath = os.path.join(directory, name + ".vcd")
+            vcdpath = os.path.join(directory, filename + ".vcd")
             if path.exists(vcdpath):
                 backup = vcdpath + '.' + str(path.getmtime(vcdpath))
                 shutil.copyfile(vcdpath, backup)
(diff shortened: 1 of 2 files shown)


cfelton:
@punkkeks The `.name` attribute is used for this.  It is the same in `traceSignals`, `toVerilog`, and `toVHDL`.

myhdl/_traceSignals.py
```diff
@@ -89,8 +91,13 @@ def __call__(self, dut, *args, **kwargs):
             else:
                 directory = self.directory
 
+            if self.filename is None:
+                filename = self.name
+            else
```
cfelton: missing the colon, I don't understand the description why the `name` attribute can't be use.

punkkeks:
No, name and filename is different, because, as I said, myhdl uses name to extract and create the hierachy in the vcd file. An example (the upper row is made with name, the lower one with filename)
screenshot from 2016-01-30 21-34-34

See https://github.com/jandecaluwe/myhdl/blob/master/myhdl/_traceSignals.py#L92

nicmcd:
I agree with @punkkeks. I've been using the `name` attribute to adjust the filename, but what I really want is to adjust the filename, not the top level instance name within the VCD file.

nicmcd:
@punkkeks, I think you should modify `myhdl/test/core/test_traceSignals.py` to reflect the desired changes.

cfelton:
I understand and agree (now) but for me I never noticed or thought it was an issue that the top-level name in the VCD hierarchy changed (match the filename).  This change will affect a lot of existing code. 

I wonder if the `.name` attribute will be used if there is a `filename` attribute?

punkkeks:
@cfelton This change may affect some code, but won't change any behavoir, because when you don't set it, it will use the original name (have you seen the diff already?!)
@nicmcd I don't know :/ As far as I can see, the test is really not a good one (at least that one for setdirectory), because it relys on internal behavoir (top.__name__ and fun.__name__) and when you change th

punkkeks:
Darn you touchpad! I wasn't ment to close this PR.

cfelton:
@punkkeks ok, that make sense and sounds reasonable.

cfelton:
@punkkeks the code looks incomplete, I see where `filename` will default to `name` if it is not set but then `filename` is not used.  The `vcdpath` is still the combination of `directory` and `name`.

punkkeks:
D'oh! Thanks for the hint. I was comiiting before thinking... In the first commit it was right. (4 commits, for such simple fix? I should go to sleep :/ )

jandecaluwe:
Thanks.

Issue #123, opened 2015-08-21
[bug] Inconsistent Hierarchy returning list of generators


cfelton:
A subtle error occurs, in some scenarios, when a list of generators is returned.  The following captures the slight differences

```python
def mod1():
    # ...
    return gen1

def mod2():
    # ...
    return [gen2]

def top():
    mod_insts = []
    mod_insts += [mod1]   # this works
    mod_insts += mod2     # this doesn't work 
```

This appears to only occur when a single generator is being returned.  [This test](https://github.com/cfelton/myhdl/blob/hier_list_error/myhdl/test/bugs/test_issue_123_2.py) demonstrates the error.

jandecaluwe:
Mm, I fear this is not straightforward.

The problem is that hierarchy extraction expects the whole hierarchy tree to be reconstructable from the return values. It does look at object ids for that purpose. But when you extend a list with another one, the latter object is gone. So the return value of mod2 is no longer found in the return value of the higher level module that instantiates it. In hardware terms, list.extend flattens out the hierarchy.

At this point, my best suggestion is to make the error message clearer. There are a couple of ways to get an "inconsistent" hierarchy; the user should get extensive feedback on what may cause it. And suggestions, like: use list.append instead of list.extend to built instance return values.

BTW: this is a good task for a student/beginner: find/list all the different cases in which you can get hierarchy inconsistencies, and suggest a help message for users.

cfelton:
Ugh, that is a tough one ... 

When I originally reported this I took part of a day to add some additional [debug information](https://github.com/jandecaluwe/myhdl/compare/master...cfelton:issue_123_hier_list_error)  (the original effort was [here](https://github.com/jandecaluwe/myhdl/compare/master...cfelton:hier_list_error) but the git history was junk and I had to clean it up).  I took a swag at adding a `dump_hierarchy` function, it produces something like:

    Hierarchy for "fft_permute_top"
       :fft_permute_top(128) [ 80, 192]
           returns rtl_s2p, rtl_permute
         +fft_permute(192)  has 0, 2, 2
             returns rtl_permute


Also added some tests, the following is a dump for one of the **list.extend** failures:

    Hierarchy for "list_of_interfaces_top_2"
    Hierarchy Error: <gen list?> <class 'list'> (68300424) missing
    Inconsistent hierarchy - are all instances returned ?
       :list_of_interfaces_top_2(232) [656, 304, 328, 720, 488, 984, 600]
[...]

PR #147, opened 2016-02-16
MEP 114

jandecaluwe:


myhdl/_module.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+
+    stack = inspect.stack()
+    # check whether the decorator is used as a descriptor
+    if (inspect.getmodule(stack[3][0]) is myhdl._module):
+        funcrec = stack[4]
+        callerrec = stack[5]
+    else:
+        funcrec = stack[3]
+        callerrec = stack[4]
```
cfelton: This code needs to be more generic, there are cases when the stack size will vary.  See the comment in #123 for more information.

jandecaluwe:
Commit 9b5a613e55a6003b04480710a71d9d8a5a983bc2 should fix the stack size issue. Not easy to test in the regressions though, because as soon as you wrap the code the issue disappears ;-)

hgomersall:
I think this is a wonderfully useful addition - it provides the opportunity to break the tight link between the python semantics and the validity of the resultant output. I'll take a look at the current state of the code shortly.

jck:
I will fully review this code over the weekend. 

However, I have a few early suggestions:
* We could make the module decorator compulsary for everything(including simulation) to prevent ambiguity. Users can still install older versions using `pip install myhdl=0.9.0` if they are not ready to port their code.
* Simulation and conversion can be methods (similar to [uhdl](http://uhdl.readthedocs.org/en/latest/api.html#uhdl.HW.convert))
* We could release a quick 0.9.1 which contains a simple module decorator which supports the new API by wrapping the existing code path; old API would still be supported but emits warnings notifying the user to use the new API. I could easily implement this decorator using code from UHDL's `HW` class.
* We could also completely deprecate the simulation and conversion functions and force users to use the methods. This would slightly simplify simulation and conversion code since they wouldn't need error checks for code using the deprecated API. Old functions would still exist but their sole purpose would be printing deprecation warnings.

cfelton:
@jck, I am not following your last bullet, maybe you need to give a quick summary of the "methods" bullet two is referring to.  I not following what myhdl object the proposed methods are tied to?  Are they are tied to the simulation class or the function being wrapped?

Should those changes be part of MEP114 or should they be a subsequent enhancement?  What is the reason including in this changeset?  I am not suggesting good or bad but maybe those changes are separate and can be discussed separately unless there is reason to include in this changeset?

jandecaluwe:
@cfelton The methods would be on the hardware instance object. This seems like an "obviously" good solution, enabled by the new decorator that makes it possible to use classes. In that case, it would be best to implement the new API directly (to minimize the pain of backwards-incompatibility) and for me to update MEP 114 and the mep-114 branch accordingly.

@jck Inspiring stuff. I think the method approach is the right way. Also, I'm considering to use `hw` instead of `module` for the decorator name: more to the point, shorter, and less ambiguous with Python terminology.

@jck The goal of the mep-114 branch is exactly to maintain compatibility with the old API, and issue deprecation messages. But I was thinking to release this as a major 1.0 version, not as 0.9.1. My idea is that minor versions should be for updates / bug fixes that don't require a feature update of the manual. But  in this case, with backwards-incompatible changes, that will definitely be required. So in my view, 1.0 would be a transition version giving everybody the opportunity to modernize, and soon thereafter 1.1 would be a clean, smaller, modern version. I would like to have that version released before GSOC, because I think it will simplify a lot of things.

jck:
All right, in that case, the methods could be cleanly implemented without deprecation checks and the legacy functions can easily wrap around the methods after printing warnings.

josyb:
@jandecaluwe using classes (in stead of functions, see my gist: https://gist.github.com/josyb/afd84c9a06fdec77f2fd) already works fine; I just delivered about 8000 lines of VHDL code using only class-based MyHDL code. Of course the ```@myhdl / @hw``` decorator will make it look/work better. May I suggest ```@hdl```? 
@jck breaking (somebody else's) code and telling them to stick to a *to be obsoleted, frozen* 9.1 branch is not considerate. Plus some people may have a **lot** of code to port ...

jandecaluwe:
About the decorator name. `@hw` looks strangely short when I try it, `@hardware` much better. `@hdl` could also be an option perhaps, but the qualified version `@myhdl.hdl` looks a little strange perhaps compared to `@myhdl.hardware`. Feedback welcome!

hgomersall:
I like @hardware, though does this seem incongruous when the object might do more than just represent hardware? I suppose the point is that the decorator is saying - "this object represents hardware (in addition to any other python things it does)", which is quite neat and concise and the term is appropriate.

josyb:
Yes ```@myhdl.hdl``` could be too much of a good thing :)
*hardware* to me is the actual physical thing represented by schematics and a PCB
What about ```@myhdl.logic?```

hgomersall:
or @myhdl.rtl

cfelton:
@hgomersall I think `rtl` has too much baggage and is not general enough, per you "incongruous" comment.

jck:
@josyb I don't think it's inconsiderate to deprecate features on major releases as long as it is easy to use older versions, but in any case, I already agreed with Jan about preserving compatibility.

Also, I think you're misunderstanding what we mean when we talk about methods instead of functions. We're going to use methods instead of functions for simulation and conversion.

josyb:
@jck Yes I mixed up methods in MyHDL source code, what you and Jan are talking about, and methods in MyHDL hardware descriptions (what I try to do).
*Inconsiderate* is indeed too strong, I apologize (I should have looked it up first), but having to *touch up* legacy code because it no longer works with the latest software release is not something I am looking forward to.

jck:
Looks good to merge!

myhdl/_block.py
```diff
@@ -0,0 +1,250 @@
...
+    def _inferInterface(self):
+        from myhdl.conversion._analyze import _analyzeTopFunc
+        intf = _analyzeTopFunc(self.mod.modfunc, *self.args, **self.kwargs)
+        self.argnames = intf.argnames
+        self.argdict = intf.argdict
+
+    # Public methods
+    # The puropse now is to define the API, optimizations later
+
+    def verifyConversion(self):
```
cfelton: whey do these need long names, wouldn't `verify` and `analyze` be sufficient?  And if not, should it follow the same naming pattern as `config_sim`?

jandecaluwe: I though that the short names were not clear enough. Of course, the naming pattern should be consistent, will change that. Given that we have `convert()` I can perhaps name them `verify_convert()` and `analyze_convert()`, a little shorter.

myhdl/_block.py
```diff
@@ -0,0 +1,250 @@
...
+
+        Args:
+            hdl (Optional[str]): Target HDL. Defaults to Verilog
+            path (Optional[str]): Destination folder. Defaults to current
+                working dir.
+            name (Optional[str]): Module and output file name. Defaults to
+                `self.mod.__name__`
+            trace(Optional[bool]): Verilog only. Whether the testbench should
+                dump all signal waveforms. Defaults to False.
+            tb (Optional[bool]): Verilog only. Specifies whether a testbench
```
cfelton: This docstring should be `no_testbench` rather than `tb`.

jck: https://github.com/jandecaluwe/myhdl/pull/157#issuecomment-196954151

zignig:
Been attempting to understand the @block decorator

How would it work for classes

@block
class thing:
  def __init__(self,a,b,c):
    self.a = a
    self.b = b
  def generate():
    @always_comb
    def thing():
      self.c = self.a + self.b
    return thing

would this work for the current impl , I like the idea of having a python class interface , rather than having to push all the signals down from the top.

jck:
I have a few suggestions regarding the implementation:

block decorator should be function based. Then, we don't need the `__get__` hack to make it work on methods. Additionally functools.wraps preserves the signature of the function/method.

```py
from functools import wraps
def block(func):
    @wraps(func)
    def deco(*args, **kwargs):
        return Block(func, *args, **kwargs)
    return deco

class Block:
    def __init__(self, func, *args, **kwargs):
        print(func, args, kwargs)
        print(func(*args, **kwargs))
        # do stuff
```

Consider a function `func` which was decorated with `@block`:
```
print(func):
old: <myhdl._block._Block object at 0x7fd0dcb8f080>
new: <function func at 0x7f1a10ba4bf8>

print(func()):
old: <myhdl._block._BlockInstance object at 0x7fc05b073b38> #Instance of _BlockInstance class
new: <__main__.Block object at 0x7f97d2559eb8> #Instance of Block class
```

jck:
NOTE: I can implement the above suggestion either before or after the merge since it is an internals change and the API stays the same.

josyb:
@zignig IMO the ```@block``` decorator needs to be applied to a function/method that does the *hdl* work, not to the encapsulating class:
```python
class thing:
    def init(self,a,b,c, d):
        self.a = a
        self.b = b
        self.c = c
        self.d = d

    @block
    def generate():
        @always_comb
        def thing1():
            self.c = self.a + self.b

        @always_comb
        def thing2():
            self.d = self.a - self.b

        return thing1, thing2
```

zignig:
Ahhh , Ok.

That's was what I was confused about, So if you have a set of various class instances with signals and @block generators myhdl should be able to extract all the "hardware" stanzas.

That makes more sense.

cfelton:
@zignig yes with limitations, the interface sections on the manual and MEP might help shed light on the limitations for conversion (convertible subset).
  * [interfaces MEP](http://dev.myhdl.org/meps/mep-107.html)
  * [interfaces what's new](http://docs.myhdl.org/en/stable/whatsnew/0.9.html#interfaces-conversion-of-attribute-accesses)
  * [interfaces conversion](http://docs.myhdl.org/en/stable/manual/conversion.html#conversion-of-interfaces)

The block-methods will be used almost identical to the block-functions

```python
def my_top_block():
    my_interface = MyInterface()
    func_block_inst = my_func_block(my_interface)
    meth_block_inst = my_interface.adapter()
    return func_block_inst, meth_block_inst
```

I have found the use of classes to represent interfaces is very powerful, 
using the block-functions is best for encapsulating the functionality
of a physical hardware block.

jandecaluwe:
@jck About a function-based implementation of `block` perhaps, but note that the current `_Block` class also maintains state. It may be possible to do this differently of course, but I propose to investigate this after the merge.

jandecaluwe:
@jck IAs I also commented in gitter, I decided to re-implement the block decorator already using a functional approach, as per your suggestion. It simplifies the code considerably, because indeed functions are descriptors already and also because it permits to do away with one level of classes, so that not we only need `_Block`, giving "instances" a consistent meaning across Python and hardware.

jck:
Awesome!

jandecaluwe:
Python lovers may appreciate the sophistication that the Block object constructor contains a reference to the decorator function that returns it, to make attributes set on the original functio available.

Issue #105, opened 2015-07-16
[1.0 feature] Initial values


jandecaluwe:
**Issue** Enable initial values for signals and variables

**Task** It's time to enable initial value support. Use `no_initial_values` as a parameter to `toVHDL()` and `toVerilog()` to disable/enable this. Default should be enabled.

**Guideline** This is best done in atomic well-defined parts. Start with the simple case of non-composite VHDL signalls & Verilog regs.

**See also**  http://dev.myhdl.org/tasks/initial-values-support.html

**Complexity estimation** Medium

Famguy:
Hello! I would like to work on this. Could anyone please tell me more about it?

cfelton:
@Famguy have you reviewed the [see also](http://dev.myhdl.org/tasks/initial-values-support.html) link above?

Famguy:
Yes @cfelton I did go through it. From what I gathered from the feature description, we need to add a support for initial signal values in the Verilog/VHDL scripts generated (for eg: the corresponding Verilog syntax being, `input d = 4'b1010`). This would involve modifying the `_toVerilog.py` and `_toVHDL.py` files to set the initial values for all inputs/wires/reg to a default value (0 maybe) or user defined ones in the generated scripts. Please do correct me if I'm wrong.

cfelton:
@Famguy yes, that is the general approach.  You can look at [this](https://github.com/jandecaluwe/myhdl/compare/master...cfelton:initial_value_support) to see the minimal changes to get started. 

One of the first things you want to do is create a test to verify initial values.  You should be able to user the [convertible testbench](http://docs.myhdl.org/en/stable/whatsnew/0.6.html#conversion-output-verification) approach (tests in conversion/general).  Once you have a test you can start working on the implementation.  These two links should help with the test creation: [contributing ](http://dev.myhdl.org/guide/guide.html) and [writing tests](http://dev.myhdl.org/guide/guide_tests.html).

Once you have a test we can discuss the specifics of the initial values, if needed.

cfelton:
@Famguy also see this issue, comments on contributing for 1.0 https://github.com/jandecaluwe/myhdl/issues/110

minghuadev:
The original "see also" link is broken. And I don't see how this feature would work. Besides, looked at cfelton's example page on bitbucket. 

The question on stackoverflow https://stackoverflow.com/questions/44922574/python-myhdl-package-how-to-generate-verilog-intial-block .

[...]

PR #150, opened 2016-02-26
Initial value support

hgomersall:
Initial value support, with tests that fit into the existing testing infrastructure.

I can confirm the iverilog works fine, but I can't get ghdl installed, so we'll need to see what travis reports.


cfelton:
@hgomersall murphy's law, it looks like the `ghdl` target did fail.

cfelton:
@hgomersall per previous comments the toV* attribute should be `no_initial_value` to be consistent with the other conversion attributes.

hgomersall:
Right, let's try again. Fix for the python 3 print error, changed the name as suggested and actually enabled the initial value case in the tests for VHDL (it was previously not set).

Let's see what happens...

hgomersall:
Hmmm still failing. Does GHDL have support for initial values?

cfelton:
@hgomersall yes, it should.

hgomersall:
Ok, I need to get ghdl installed. Bear with me, it might take a few months.

cfelton:
@hgomersall it fails with Modelsim as well (`vcom`)

hgomersall:
I think it should pass with modelsim now. I can manually confirm the output is consistent when using Vivado.

The problem with GHDL is that by default it doesn't include std_logic_textio, which is used to reliably print arbitrary width vectors. It can be enabled with "--ieee=synopsys". I see no reason why we don't add this to the simulator, but that's quite a far reaching change to satisfy a test, so I'm proposing it before making the changes in the PR. Any comments?

cfelton:
@hgomersall yes it passes with Modelsim (`vcom`).  

I guess my thought is, remove the need for the arbitrary large vectors in the test.  The [manual](http://docs.myhdl.org/en/stable/manual/conversion.html#known-issues) does call out some conversion limitations, we could add the GHDL limitation in the docs.  And then create a new issue and test for the arbitrary long vectors.  The long vector specific test can be marked with the `pytest.mark.xfail` (and there is even an option to only specify xfail with simulator type ghdl) and then it can be re-enabled when that issues is fixed.

This should allow us to move forward with the first part of the IVS support (#105) and still track the issue encountered.  And we will also need to decided if list-of-signals IVS is part of #105 or a separate issue/feature should be created.

hgomersall:
IVS doesn't actually fail with GHDL, just the test for it. This is simply down to the problem of vanilla VHDL not being able to print vectors that can't be converted to 32 bit.

hgomersall:
Oh, in which case surely the solution is just to replicate the code...

hgomersall:
Except that doesn't work because it's dumped into the code as a string from the python end.

cfelton:
@hgomersall I don’t understand your terse comments?  Why would you have to duplicate code?  If you don’t want to put the `test_long_signals` in a separate test file you could at least mark it as `xfail`

```python
failing_simulator = _simulators[‘ghdl’]
reason = """GHDL doesn’t include std_logic_textio by default and 
long vectors fail to print correctly during post conversion simulation
output comparison"""

@pytest.mark.xfail(verify.simuator == failing_simulator, reason=reason)
def test_long_signals():
```

In my opinion, this doesn’t hold up the IVS efforts and the bug/issue is still recorded.  I think a separate test and creating an issue (then `@bug` could be used) would be best but a minor preference.

jandecaluwe:
--ieee=synopsys is a no-go. See http://home.gna.org/ghdl/ghdl/IEEE-library-pitfalls.html

hgomersall:
@jandecaluwe fair enough. Perhaps we could extract the suitable bits from: https://github.com/tgingold/ghdl/blob/master/libraries/synopsys/std_logic_textio.vhdl
Alternatively, do you have a better suggestion as to how to compare long vectors? I feel such a comparison is important in completing the tests - I've come across a couple of conversion bugs related to VHDL's handling of ints.

@cfelton Apologies for the terseness - one should probably give the response one's full attention rather than trying to quickly respond in between things. I concur with your position. There seems little reason to wait on the final test being complete for GHDL before merging - the code is actually well tested in that scenario and I'm satisfied there is no problem.

jandecaluwe:
As this all seems painful, I reviewed some options. In principle, I think VHDL 2008 should make things a lot easier. It seems that ghdl 0.33, the latest release, has better support for VHDL 2008. However, this is a version that requires a custom install on travis. Today, I have made and committed the changes required to support this. The test suite now passes with ghdl 0.33, with VHDL 2008 enabled.

I don't know yet whether VHDL 2008 support in ghdl is sufficient to simplify the issues enountered here. I will try that next. I propose to put things on hold until then.

hgomersall:
@jandecaluwe great, thanks!

jandecaluwe:
VHDL 2008 support has been a success so far. It required a custom install of the latest version of ghdl (0.33), but from then on the new features worked as expected.

The result is a drastic simplification of printing in VHDL, but also some significant implications. In particular, I have switched to a hex representation when printing vectors, to be able to support arbitrary long vectors. This has some ramifications in other parts of the code - the test suite passes but there may be issues. The latest public commit has all my changes so far.

I propose to reconsider this PR from the ground up. I recommend to investigate whether a (signed) hex representation can be used for initial values on vectors like signed/unsigned, as VHDL 2008 seems to promise. This info will be useful to switch from verbose binary to hex literals later in other parts of the code also.

A local installation of ghdl 0.33 will be mandatory.

cfelton:
@hgomersall we have a GSoC student that would like to assist with this PR: @udara28.  He completed a project last year and is applying for another this year.  If there is something he can do to help on this PR feel free to let him know.  

@jandecaluwe When you say "reconsider from the ground-up" do you simply mean, step back modify the long bit-vector tests and then resubmit as a new PR (merged with the latest VHDL 2008 changes) or do you mean something greater?

jandecaluwe:
@cfelton A PR is a living thing. It is really a public branch, and as long it is not merged you can keep on modifying/improving it. No need to raise a new PR unless you want to throw everything way or need to start all over again.

With "reconsider from the ground-up" I mean to make best use of VHDL 2008 features on all aspects but *within the context of this PR*. In particular, not just fix the tests, but also check whether initial values themselves can be represented in a better way than as long string of 0/1s. I can only repeat: I recommend to investigate whether a (signed) hex representation can be used for initial values on vectors like signed/unsigned, as VHDL 2008 seems to promise. This info will be useful to switch from verbose binary to hex literals later in other parts of the code also.

cfelton:
@jandecaluwe perfect, thanks for the clarification.  

@udara28, using git you should pull from @hgomersall `initial_value_support` branch.  Then you will be able to create a pull-request to @hgomersall with your changes.  Let us know if you have questions after reviewing the current changes and this PR comment thread.

hgomersall:
@cfelton Apologies for the delay on this (and other issues) - I've been away for a week. I'll respond properly on Wed but having a GSoC student would be great.

@udara28 It should be relatively simple to modify for VHDL 2008, so please crack on.

hgomersall:
The current implementation passes tests and is expected to work.

Do we want to allow the convertible code to depend on VHDL 2008? The current IVS implementation uses old style vector literals (`:= "1111100"`). IMO this is better for backwards compatibility but is much more verbose than can be achieved with the 2008 enhanced hex literals.

jandecaluwe:
Now that ghdl has VHDL 2008 support, I have no indication that there is a problem with requiring VHDL 2008 support from external tools and making full use of it. We can't wait on commercial software forever ;-)

One complication is that @jck has explicitly asked me to wait with changing/adding tests until he has done his thing with the test framework. I will honour that, afterwards a modification to the test in this PR may be required also.

hgomersall:
@jandecaluwe what is the vhd_int type used for in toVHDL? All the intbvs I am converting yield vhd_signed or vhd_unsigned - should I ever expect a convertible type to be vhd_int?

myhdl/conversion/_toVHDL.py
```diff
@@ -113,6 +113,7 @@ class _ToVHDLConvertor(object):
                  "use_clauses",
                  "architecture",
                  "std_logic_ports",
+                 "no_initial_value"
```
jck: I suggest renaming this to `initial_values`

hgomersall: @jck Do you mean `no_initial_values` or `initial_values`? I've no strong opinion so happy to take guidance.

jck: I mean, get rid of the 'no' since it is counter intuitive. I suggested plural for consistency.

hgomersall: I'm easy with that, though I'm not sure there wasn't some thinking about default consistency in the long term. I think the plan is to ultimately make it default `False` and then `no_initial_values` makes sense.

hgomersall:
The latest update uses the 2008 VHDL hex literals. I have a few questions regarding the implementation however:

1. I noticed that `s._init` within the siglist passed to `_writeSigDecls` already contains a suitably hexified twos-complement init value. Where does this get set and is it robust to use? (certainly, the `_init` of the signal when it's created prints a signed int).
2. Do we ever need to handle a `vhd_int` type signal? I've yet to see this appear, so perhaps someone could advise when this might occur.
3. I think using a python generated full-size hex representation is simpler than VHDL 2008's signed and unsigned literal forms. The problem with these is there will be extra logic for the positive and negative cases with some unnecessary coercion just to present the twos complement negative values in the right form.

hgomersall:
Oh, also I change `no_initial_value` to `no_initial_values` which seemed to make more sense.

Nicolasp38:
VHDL 2008 in not much used by FPGA vendors. To my knowledge, the last Xilinx implementation tool (Vivado) does not make use of VHDL 2008. One have to enable it manually by a special command.
Why not use an option for VHDL 2008 use ?

hgomersall:
@Nicolasp38 It's easy enough to enable in Vivado and the result is correct.

Nicolasp38:
@hgomersall Sure but :
- If you are not aware of this "detail", you can think VHDL generated by MyHDL is broken.
- You can't use an "old" implementation tool (ISE) for "old" series which are still commonly used (Spartan6).

hgomersall:
@Nicolasp38 the only way I could see this working would be to disable features if VHDL 2008 is disabled.

hgomersall:
Updated to reflect the gitter decision to use `initial_values`.

hgomersall:
The latest commit highlights the problem raised in #165 . That problem is worked around rather badly in the test suite using custom vhdl and verilog code, and printing `.val` explicitly in myhdl.

hgomersall:
The recent changes are about supporting initial values on list of signals, allowing (among other things) memory initialisation.

hgomersall:
This currently doesn't work for ShadowSignals. Is that a thing we want?

hgomersall:
What needs to be done to have this merged? Are there still unknowns?

cfelton:
@hgomersall you might want to ping @jandecaluwe on gitter.

PR #152, opened 2016-03-03
updated doc for the new public method sim.quit()

srivatsansoft:
I have added the line sim.quit() at places where there was a chance of creating more than one simulation instance.
At places where the simulation was run for a specified duration, but only one instance of simulation was created in the program I didn't add sim.quit() line as the program was anyways going to get terminated.


jandecaluwe:
Nice work - documentation is important!

PR #155, opened 2016-03-10
Increasing test coverage for tristate and delay

Vikram9866:
In call to super changed function to object `Tristate` replaced by `_Tristate`
Instead of
    `super(Tristate, self).__init__(val)`
It should be
   `super(_Tristate, self).__init__(val)`

Reasons:
  Tristate is a function and _Tristate is a class.
  super function is defined as :
  `super(type, obj)` -> bound super object; requires `isinstance(obj, type)`
  `super(type)` -> unbound super object
  `super(type, type2)` -> bound super object; requires `issubclass(type2, type)`

 and `super(Tristate, self)` does not lie in the above mentioned syntax.

Initial test coverage:
>|Name|Stmts|Miss|Cover|
>|---|---|---|---|
>|myhdl\_delay.py|9|1|89%|
>|myhdl\_tristate.py|55|36|35%|

Increased test coverage:
>|Name|Stmts|Miss|Cover|
>|---|---|---|---|
>|myhdl/_delay.py|9|0|100%|
>|myhdl/_tristate.py|55|10|82%|

### DIFF
--- a/myhdl/_util.py
+++ b/myhdl/_util.py
@@ -23,6 +23,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import __future__
 import ast
 import sys
 import inspect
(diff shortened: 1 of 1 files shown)


myhdl/_tristate.py
```diff
@@ -23,7 +23,7 @@ class _Tristate(_Signal):
             
     def __init__(self, val):
         self._drivers = []
-        super(Tristate, self).__init__(val)
+        super(_Tristate, self).__init__(val)
```
jck: Good catch, I think this was definitely a bug.

jck:
I don't understand the point of these delay tests, can you explain?

Vikram9866:
I added delay test to increase test coverage for `_delay.py `. Its a very trivial check to make, but I assume having `100%` test coverage would be better.

jck:
I mean, what exactly is it testing for?

Vikram9866:
There was no test case checking the behavior of `delay` function for non-integer and negative delay values. So I made a separate test file which verifies the expected behavior.

`delay` function present in `_delay.py` should raise a `TypeError` for negative and non-integer cases.

PR #157, opened 2016-03-15
Mep 114 update

jck:


jck:
Obviously, the `convert` API isn't set in stone yet. As of now, it allows one to configure all the attributes of the converters.
However, we can improve this API. We don't need to stick with the attributes present in the present converter classes. For example, the `convert()` method supports a `tb` optional argument which can be used instead of `no_testbench`(which is counter intuitive). We could also make additional improvements, such as allowing `path` to be a file instead of a directory and inferring the HDL based on the extension.

PR #158, opened 2016-03-15
docs: custom directive for including/running examples

jck:
This adds two custom directives (include-example, run-example) which includes or runs a file from example/manual. This PR uses the directive only for hello1.py just to make sure nothing goes wrong with readthedocs.


jck:
As I feared, this breaks the docs build.
In https://readthedocs.org/dashboard/myhdl/advanced/ , you must enable the `Install your project inside a virtualenv using setup.py install` checkbox.

jandecaluwe:
Seems to work now!

PR #159, opened 2016-03-19
Increasing test coverage for encoding in enum

atharvaw:
```
                  Name                                       Stmts   Miss  Cover
```

Initial               myhdl_enum.py                         112     37    67%
Increased       myhdl_enum.py                         112     34    70%

The test coverage for enum encoding, enum length, enum string type and enum copy is added .


atharvaw:
Can you please help why Travis CI build failed failed.

cfelton:
@atharvaw The myhdl.enum encapsulates the encoding value, you can't compare directly an int to the enum type.  Verifying explicitly the values to be generated in myhdl simulation can't be done.  That specific test would need to be part of the conversion test.

It still could be worthwhile to generate enums with the encoding and verify they can be used.  This could still be worthwhile to increase the coverage in the core tests.

```python
def walk_states(encoding):
    states = enum('search', 'confirm', 'sync', encoding=encoding)
    state = Signal(states.search)
    @instance
    def tbstim():
        state.next = states.confirm
        yield delay(2)
        state.next = states.sync
        yield delay(2)
        state.next = states.search
    return tbstim


def test_enum_onehot():
    Simulation(walk_states('one_hot')).run()
```

I assume the `one_hot` and `one_cold` paths weren't tested in core?  A test like this could be useful to quickly catch simple errors in the path.  But don't make the too complicated, try to keep is simple and add comments in the code what you are testing.

atharvaw:
Thanks for your review, I will look into it.

cfelton:
Here is the coverage when you include the conversion verilog testing (it will be higher when you throw in the VHDL)

```
myhdl/myhdl/_enum.py                         114     25    78%   
   source lines not tested:
   32, 36, 45, 56, 89, 92, 100, 102, 106, 109, 115, 121, 
   123, 128, 130, 150, 153, 164-171
```
Note 164-171 would be covered with the VHDL target testing.

There are a couple error paths that are test.  Example, if you pass the wrong encoding type, you would be able to test this with `with pytest.raises(ValueError)`.   Many of the missing should be small little tests to verify.

atharvaw:
I have added test coverage for enum encoding, length, string type and copy.

myhdl/test/core/test_ShadowSignal.py
```diff
@@ -165,8 +166,9 @@ def check():
         yield delay(10)
         assert s == c
         c.next = None
+        d.next = 0
         yield delay(10)
-        assert s == None
+        assert s == d
```
cfelton: Why were these changes made?  You should back out these changes since the PR is specific to expanding the enum test coverage.

cfelton:
@atharvaw update the description with the latest test coverage and also add the addition checks.  It will all be summarized in the PR description and folks will not need to read through the comments.

Issue #95, opened 2015-07-07
[enhancement] toVHDL name collision checks


cfelton:
Users often run into the scenario when the converted VHDL is invalid because of name collisions.  This typically occurs because VHDL is not case sensitive and Python is.  The toVHDL converter should produce warnings for signal, variable, and port names that conflict and/or are invalid (e.g. dash in the name '-' or starting with an underscore '_').

Example of an incorrect conversion:

```python
def pipe(xi, xo, clock, reset, N=8):

    buffer = [Signal(xi.val) for _ in range(N)]

    @always_seq(clock.posedge, reset=reset)
    def rtl():
        buffer[0].next = xi
        for n in range(1, N):
            buffer[n].next = buffer[n-1]
        xo.next = buffer[N-1]

    return rtl
```

[complete code example](https://gist.github.com/cfelton/cddeebe2f96b00b53ed5)

hgomersall:
Should this include reserved words? Or is that a different bug?

cfelton:
In this original enhancement suggestion I included reserved words, they could be separated if they are easier to resolve one by one.  This might have been addressed in #160.

PR #160, opened 2016-03-25
[RDY] Name collision fix (Issue #95)

MrCanadianMenace:
Added new .py file to save all VHDL reserved words and variable/signal names to raise a warning if a variable/signal is created with a reserved word or recycled name.

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -354,7 +354,7 @@ def _writeModuleHeader(f, intf, needPck, lib, arch, useClauses, doc, stdLogicPor
             if convertPort:
                 pt = "std_logic_vector"
              # Check if VHDL keyword or reused name
-            _nameValid(s)
+            _nameValid(s._name)
             if s._driven:
                 if s._read:
                     if not isinstance(s, _TristateSignal):
@@ -419,7 +419,7 @@ def _writeSigDecls(f, intf, siglist, memlist):
             print("signal %s: %s%s;" % (s._name, p, r), file=f)
         elif s._read:
             # Check if VHDL keyword or reused name
-            _nameValid(s)
+            _nameValid(s._name)
             # the original exception
             # raise ToVHDLError(_error.UndrivenSignal, s._name)
             # changed to a warning and a continuous assignment to a wire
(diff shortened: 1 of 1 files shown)


myhdl/conversion/_VHDLNameValidation.py
```diff
@@ -0,0 +1,47 @@
...
+                  "wait", "when", "while", "with", "xnor", "xor"];
+
+    #A list to hold all signal names being used in lowercase to raise an error
+    #if no names are reused with different casing
+    _usedNames = [];
+
+    #Function which compares current parsed signal/entity to all keywords to
+    #ensure reserved words are not being used for the wrong purpose
+    def _nameValid(keyword):
+        for i in _nameCheck._vhdl_keywords:
```
cfelton: This code is incorrect, it should be something like:

```python
def _nameValid(name):
    for keyword in _nameCheck._vhdl_keywords:
        if name == keyword:
```

The value passed to the checking function is not the keyword but rather the parsed strings/names.

MrCanadianMenace:
I added the recommended changes and changed the rest of the new code to match the naming which was recommended

nturley:
I wrote some name collision tests when I was experimenting with extended identifiers.

It's here: https://github.com/nturley/myhdl/blob/65c976a22d258ea7f81c3e808a016e8308056382/myhdl/test/bugs/test_collision.py

I was in a silly mood when I wrote them and quoted the song "Collide" and made two signals named you_and_i collide.

MrCanadianMenace:
Gotta have some fun with programming right?  Not entirely sure of where I should go with this now though, is it preferred to have this code setup more independent from the code conversion or should I keep the current style and do some more cleaning and refining for a merge?

MrCanadianMenace:
Fixed the main problems I was having but am not entirely understanding the travis-ci build errors, any guidance here would be very much appreciated

jck:
Try running the tests on your local machine in a clean virtualenv. You'll get the same errors you see in the travis web ui. If that's not the case, let me know since that means that there's something wrong with travis or our setup.

MrCanadianMenace:
@cfelton got tests passing and was just wondering if there was anything else you think should be changed

PR #163, opened 2016-04-14
Small fix to enable self.name be used in exceptions in _verifySubs in _Block

hgomersall:
Prior to this patch, `_verifySubs` is called before `self.name` is assigned, but in some exceptions `_verifySubs` uses `self.name`, so we get a fairly unhelpful attribute error about missing `name` rather than useful information about the actual error.


hgomersall:
Added to the error message that checks what is returned from a block to inform the user which block is causing the problem (helpful when there are lots of blocks in a hierarchy).

PR #164, opened 2016-04-14
[WIP] Improved block decorator

hgomersall:
Consider the following code:

    from myhdl import *

    @block
    def drive(clock, a, b):
        @always(clock.posedge)
        def driver():
            a.next = 2
            b.next = 2
                
        return driver

    @block
    def clock_gen(clk):

        @always(delay(10))
        def driveClk():
            clk.next = not clk

        return driveClk

    class Gah(object):

        def __init__(self, clock):
            self.clock = clock

        @block
        def foo(self, a, b):
            driver = drive(self.clock, a, b)
            clock_driver = clock_gen(self.clock)
            return [driver, clock_driver]

    clock = Signal(False)

    a,b = Signal(intbv(0, min=-10, max=10)), Signal(intbv(0, min=-10, max=10))

    gah = Gah(clock)
    gah.foo(a, b).convert(hdl='VHDL')

As currently implemented in master, the block decorator will not be able to handle the extra `self` argument to the top level conversion function `gah.foo`. This is due to the differences in the way instance methods should be handled compared to normal functions by decorators.

The result of running the above is an inclusion of an extra interface port from `clock`, which it then complains about being read internally. Clearly `clock` should not feature in the interface at all. This arises because `self` is treated as an object to inspect as being part of the interface, so its name space is checked for Signals (AFAICT, `self` is assumed to be an interface object).

This issue of decorators and their subtleties discussed at some length in the following blog posts:
http://blog.dscpl.com.au/2014/01/how-you-implemented-your-python.html
http://blog.dscpl.com.au/2014/01/the-interaction-between-decorators-and.html

This PR fixes the problem with the approach highlighted in the second post. 

It is a WIP because it doesn't complete the decorator to the case described in the second post using transparent proxies (which may be overkill), and also because one `conversion.test_hdlobjnotself` now fails. The test failure looks like it is directly related to this issue, so more advice is needed (I haven't yet looked at the test failure with a view to debugging).

More useful decorator related posts from the above author can be found here:
http://blog.dscpl.com.au/p/decorators-and-monkey-patching.html

### DIFF
--- a/myhdl/_block.py
+++ b/myhdl/_block.py
@@ -22,7 +22,8 @@ from __future__ import absolute_import
 
 import inspect
 
-from functools import wraps
+#from functools import wraps
+import functools
 
 import myhdl
 from myhdl import BlockError, BlockInstanceError, Cosimulation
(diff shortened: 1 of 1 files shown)


hgomersall:
This patch builds implicitly on #163.

jandecaluwe:
I see the point, or at least I think I do. Good digging work, thanks.

The failing top-level test actually revealed a real issue that still had to be addressed with the new implementation. I have fixed that now.

I don't think we have to worry about performance, as the block calls are at elaboration tim.

hgomersall:
I concur re performance. Thanks.

jck:
I'm re-examining the current implementation of the block decorator and I have a few questions.

I'm having a hard time understanding your example. Irrespective of the block decorator, it wouldn't convert anyway because of the `@always(delay(10))`

Could you provide another example which would have the issue you're describing?

jck:
FWIW, reverting this PR and going back to the `functools.wraps` approach fixes #209 and the new tests you introduced in #189.

hgomersall:
Sorry, what have you done? I'm really confused...

`functools.wraps` can't be used for the reasons outlined in links in the initial PR. I agree the implementation is not obvious, but is the "correct" way to do it. The subsequent PRs deal with resultant problems that I've encountered (and it's really quite well exercised now).

Basically, `functools.wraps` is too inflexible for doing anything like what we need it to do. The original problem was not #209, it was the problem as described in this PR description.

It's easy to fix the conversion (not quite sure why I wrote it so wrong), just replace the clock source with a convertible clock source:
    @instance
    def driveClk():

        while True:
            yield(delay(10))
            clk.next = not clk

The point is (unless something else has changed) the self is not handled properly using `functools.wraps`. This can't be worked around without a proper implementation of a class/function agnostic decorator. It took a while for me to get my head around, but the links explain it all.

jck:
Apologies, I think I've fundamentally misunderstood the decorator issues. I'll go through those links more carefully.

hgomersall:
:smile: no probs!

Issue #169, opened 2016-05-24
Naming Conversion Conflict


srivatsan-ramesh:
[Here](https://gist.github.com/srivatsan-ramesh/ed9b2106b42c3a6f9f82a8bc86b5454e
), in this example there are two instances, one of `Test1.test()` and another `Test2.test()` , both these instances will have the same name` test_0 `. So, when this code is converted it produces two wires with the same name for the `clock` signals.
`wire test_0_self_clock;`
`wire test_0_self_clock;`
Due to this, when signals are being traced, the waveform obtained is not of two `clock` signals but of a concatenation of the two `clock` signals.

PR #170, opened 2016-05-24
Added test for issue_169

srivatsan-ramesh:
Added a test file which fails due to the error described in #169


cfelton:
@srivatsan-ramesh thanks for the test addition.

@jandecaluwe I had instructed @srivatsan-ramesh to put this test in the `bugs` but this is for unreleased code.  If we decide to include the test, should this test stay in `bugs` or move to `test/conversion/general`?

PR #175, opened 2016-06-03
Added nested interface test

cfelton:
It was pointed out that the interface tests defined a nested interface but did not use it.  The test has been updated to use the nested interface.

Additional changes.  General cleanup to the interface test code.  Added the `-2008` switch for the `vcom` simulation target.  Changed the tests to use `analyze_convert` and `verify_convert` methods.


myhdl/conversion/_verify.py
```diff
@@ -69,7 +69,7 @@ def registerSimulator(name=None, hdl=None, analyze=None, elaborate=None, simulat
 registerSimulator(
     name="vcom",
     hdl="VHDL",
-    analyze="vcom -work work_vcom pck_myhdl_%(version)s.vhd %(topname)s.vhd",
+    analyze="vcom -2008 -work work_vcom pck_myhdl_%(version)s.vhd %(topname)s.vhd",
```
jck: This is a pretty big change which would affect all conversion tests. Should we document it first?

cfelton: @jck I believe I added this because there were other changes or discussion that eluded.  We can make this a separate pull-request if that simplify things.

cfelton:
merged with the latest master.

Issue #125, opened 2015-09-02
Usability of Interfaces


josyb:
Yesterday I decided to try using an Interface. 
I am currently experimenting with a StructType object to mimic the VHDL _record_ / SystemVerilog _struct_, and wanted to get a feel for the 'how and what' of the Interface.  
Some background: I use MyHDL to develop re-usable modules for use with Altera's Qsys. Qsys has two main interfaces, the Memory Mapped (MM) and the Streaming (ST) types. All of the modules have (at least) one input and (at least) one output ST-interface. So the ST was a good candidate for my first try.
I also have a utility to generate the necessary _hw.tcl file for Qsys to be able to recognise/parametrise the module. This utility also steers the conversion.
I added some code to generate the members for the ST-interfaces on demand.

I was kindly surprised, because it looked good. The ports were all there and had the correct names matching the advertised names in the corresponding _hw.tcl file.

So today I decided I thus had to upgrade all my existing modules to conform to my latest changes in the Qsys utility, not necessarily because of the interfaces but in anticipating using interfaces I had already changed some port naming rules.
The first module mentioned above was a rather simple one and only had one level of functions, better said the top function was the only RTL function.
Now when trying/ working on a larger module where the top function only collects the RTL from lower modules I find that the names are 'mangled'.
This MyHDL code
```python
import myhdl

class fi( object ):
    ''' a simple interface '''
    def __init__(self):
        self.Data = myhdl.Signal( myhdl.intbv()[8:])
        self.Ready = myhdl.Signal( bool(0) )
        self.Valid = myhdl.Signal( bool(0) )
        
                
def mangler( Clk, Reset, D, Q):
    ''' a silly module to mangle the input data '''
    @myhdl.always_comb
    def mangler():
        Q.Data.next[:4] = D.Data[4:]
        Q.Data.next[4:] = D.Data[:4]
        Q.Valid.next = D.Valid
        D.Ready.next = Q.Ready
    
    
    return mangler


def topfi(Clk, Reset, D, Q):
    ''' the top level '''
    
    mnglr = mangler(Clk, Reset, D, Q)
    
    return mnglr


def convert():
    
    Clk = myhdl.Signal( bool(0) )
    Reset = myhdl.ResetSignal(0, active=1, async=True)
    D = fi()
    Q = fi()
    
    myhdl.toVHDL( topfi, Clk, Reset, D, Q)
    

if __name__ == '__main__':
    ''' '''
    convert()
```

produces the following VHDL:
[...]

PR #177, opened 2016-06-19
Conversion of Nested Interfaces in Top Level

josyb:
My GSOC2016 *pupil* Vikram asked me once about the conversion of nested interfaces, and I had to answer it was not possible.
This PR addresses that enhancement.

It also addresses issue #125 where I posted a query on the *name mangling of interface objects*; @nturley suggested a solution which I also incorporated in this PR.
As a side effect I had to *xfail* the ```test_name_conflict_after_replace()``` in test_interfaces2.py as the renaming action now gets overwritten in the interface name expansion code.

### DIFF
--- a/myhdl/conversion/_analyze.py
+++ b/myhdl/conversion/_analyze.py
@@ -1235,29 +1235,37 @@ def isboundmethod(m):
 
 
 def _analyzeTopFunc(func, *args, **kwargs):
+    def expandinterface(v, name, obj):
+        ''' a local function to drill down to the last interface '''
+        for attr, attrobj in vars(obj).items():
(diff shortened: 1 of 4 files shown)


josyb:
The failing check above is in test_tristate,py. As PR #174 has not yet been approved, I can't run that test on my Win10 machine because I started off with a *clean* branch ...
As *test_tristate.py* is using an *interface* it may well be *upset* by the code changes.

josyb:
With some hindsight: I will split this PR into two separate ones: the first to expand nested interfaces, and the second to deal with the name mangling (#125).
So I will close this PR and delete the branch

PR #183, opened 2016-07-18
Add an interation limit when simulating to detect combinatorial loops

### LINKED ISSUE #180: Combinatorial loop hang the simulator
Creating a combinatorial loop hangs the simulator in an infinite loop. I expected it to either update at the highest possible frequency (at the tick rate) or to raise an error and abort the simulation. Test case:
```python
from __future__ import print_function

from myhdl import *

@block
def Demonstration():
    a = Signal(False)

    @instance
    def poke_loop():
        yield delay(1)
        a.next = not a

    @always(a)
    def comb_loop():
        a.next = not a

    return (comb_loop, poke_loop)

demo_inst = Demonstration()
demo_inst.config_sim(trace=True)
sim = Simulation(demo_inst)
sim.run(10)
```

rqou:
If the simulation runs for more than the specified number of ticks without advancing time, then declare the simulation stuck and abort. Fixes #180.

### DIFF
--- a/myhdl/_Simulation.py
+++ b/myhdl/_Simulation.py
@@ -112,11 +112,12 @@ class Simulation(object):
     def quit(self):
         self._finalize()
 
-    def run(self, duration=None, quiet=0):
+    def run(self, duration=None, quiet=0, iter_limit=1000):
         """ Run the simulation for some duration.
 
         duration -- specified simulation duration (default: forever)
         quiet -- don't print StopSimulation messages (default: off)
+        iter_limit -- step limit for not advancing time; used to detect combinatorial loops (default: 1000)
 
         """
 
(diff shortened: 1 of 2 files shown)


rqou:
Hrm, this PR seems to keep triggering flake in Travis, but I'm not sure why?

jandecaluwe:
Seems like a lot of code for limited usefulness. As it is in the inner simulation loop, I am especially afraid it will affect performance. Many compiled simulators don't even have this feature. Moreover, it will not detect all cases: if you have a loop that advances time, it will not be detected.

Unless good arguments to the contrary, I am inclined to reject.

josyb:
With hindsight, and running a lot of lengthy simulations nowadays, I overlooked the impact of this change.

PR #184, opened 2016-07-18
Remove the limitation to have only one cosimulation

rqou:
This patch removes the limitation of only one cosimulation being allowed at one time. This would be useful for e.g. running both Icarus Verilog and GHDL (which I have working but haven't cleaned up as a PR yet) as sub-simulators if a project has dependencies written in both Verilog and VHDL.

There is no unit test for this, but all existing tests (except the explicit test for multiple cosimulators) pass. I have a giant integration test that is working with these changes, but I also haven't cleaned that up to publish yet.

### DIFF
--- a/myhdl/_Cosimulation.py
+++ b/myhdl/_Cosimulation.py
@@ -34,7 +34,6 @@ _MAXLINE = 4096
 
 class _error:
     pass
-_error.MultipleCosim = "Only a single cosimulator allowed"
 _error.DuplicateSigNames = "Duplicate signal name in myhdl vpi call"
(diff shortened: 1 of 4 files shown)


rqou:
The not-cleaned-up code for GHDL cosimulation can be found at https://github.com/rqou/myhdl-ghdl-duct-tape

rqou:
Here is a demonstration that exercises this patch and uses two external simulators: https://github.com/rqou/myhdl-vhdl-verilog-test

cfelton:
@rqou, note @jandecaluwe had posted that he was going to be busy for awhile and some of others are busy with gsoc.  Don't be discourage if no one responds, we will get to it.

PR #188, opened 2016-08-04
Add vhdl_/verilog_instance to _Block decorator

josko7452:
In my workflow I have appreciated use of verilog_instance and vhdl_instance.
This PR adds support od verilog/vhdl_instance to @block decorator.


hgomersall:
This isn't actually necessary. A fairly neat workaround is to use a string-like object:

https://github.com/hgomersall/Ovenbird/blob/master/ovenbird/vivado_ip.py#L455

I'm using this in anger right now and it doesn't tread on the toes of MyHDL at all.

hgomersall:
Oh, I misunderstand this. Apologies, I thought it was related to vhdl_code. Ignore me :)

PR #189, opened 2016-08-04
[FIX] fixed a couple of issues with the way classes are handled by the new block decorator

hgomersall:
Fixed problem with setting vhdl_code and verilog_code on class methods and added test.

There was a problem with the way a unique class proxy was created every time, so changes to that proxy (e.g. setting `vhdl_code` or `verilog_code`) would be lost immediately.

The change below fixes this by always returning the same proxy given a class instance.

A test is also implemented testing the problem which works only after the fix.

Also, fixed a problem with the way class methods are named which resulted in name collisions, notably on signal names.

### DIFF
--- a/myhdl/_block.py
+++ b/myhdl/_block.py
@@ -106,6 +106,7 @@ class _bound_function_wrapper(object):
                       self.srcline, *args, **kwargs)
 
 class block(object):
+
     def __init__(self, func):
         self.srcfile = inspect.getsourcefile(func)
         self.srcline = inspect.getsourcelines(func)[0]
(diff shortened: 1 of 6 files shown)


hgomersall:
It would be great to have this merged - the conflict is pretty trivial. There is an initial value fix that I can't push until this is (compounded changes make it hard).

hgomersall:
To be clear, the resolution is to use the master version and it should be fine.

cfelton:
@hgomersall can you resolve the conflict and update this PR.

hgomersall:
Yeah, I'll do that. I was going to do it yesterday but didn't have SSH access.

hgomersall:
Once the checks pass, this should be ready to go.

hgomersall:
I've made a few more changes to this branch. The impetus was a fix for another naming bug in which we had a single class instance and multiple (>3) calls to a block method on that instance.

I think the way the names were handled previously was a little bit dodgy - the name was set on the method itself, which was then passed to the creation of `_Block`, from which the name was looked up. It worked because `_Block` then creates a copy of the name which is kept with that unique `_Block` instance. However, behind the scenes, subsequent calls to the function update the name to something new. This means `_Block` has a concept of a name (which is correct) and a reference to a function which contains some arbitrary name depending on what the last function to call it was.

Better IMO is for the function to have no concept of a name, and it to be created only on the `_Block` instance. This is what this latest patch does, and does it by adding a `name` argument to `_Block`. It makes no material difference to the public API because all the calls to `_Block` happen through wrappers (which are unchanged). It could conceivably break code that uses `_Block` directly, but since this is a dev release I see no problem with this.

All tests pass fine with the change.

josyb:
The failing tests are due to GHDL no longer functioning because of a name/directory change.
It is easy to fix, but we need somebody to merge the changes ...

josyb:
@hgomersall As @jandecaluwe has fixed the GHDL invocation you can update this PR, once more ;)

jandecaluwe:
I hate to see the complexity that methods add to the block feature. But I understand these concerns are too late here, so I reluctantly merge it in.

PR #201, opened 2016-11-07
Simulation configuration + progress callback

NicoPy:
1) After "block" instanciation, it is not more possible to configure simulation. Re-enables this possibility.
2) Add the possibility to not create trace backup.
3) Add the possibility to pass a callback to simulation run to get progress information.

### DIFF
--- a/myhdl/_traceSignals.py
+++ b/myhdl/_traceSignals.py
@@ -56,7 +56,8 @@ class _TraceSignalsClass(object):
                 "directory",
                 "filename",
                 "timescale",
-                "tracelists"
+                "tracelists",
+                "tracebackup"
                 )
 
     def __init__(self):
@@ -65,6 +66,7 @@ class _TraceSignalsClass(object):
         self.filename = None
         self.timescale = "1ns"
         self.tracelists = True
+        self.tracebackup = True
 
     def __call__(self, dut, *args, **kwargs):
         global _tracing, vcdpath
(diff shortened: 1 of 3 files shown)


NicoPy:
Sadly, Travis CI tests  failed because of network errors.

myhdl/_traceSignals.py
```diff
@@ -65,6 +66,7 @@ def __init__(self):
         self.filename = None
         self.timescale = "1ns"
         self.tracelists = True
+        self.tracebackup = True
```
jck: I think this should be False by default to avoid breaking people's workflows.

NicoPy: Which parameter are you talking about ?

NicoPy: Just realized you're talking about self.tracebackup
I added this parameter because traces are always backed-up. As I don't want/need this behavior, I added the possibility to not backup the traces and I set this parameter to True by default to mimic current behavior.

jandecaluwe:
Also here, code in the main simulation loop. I hesitate. I think I would prefer to see that part go. Is this mandatory for someone, and if so, can it be updated (also conflicts with current master)?

NicoPy:
The code in the main simulation loop just computes the simulation progress from 0% to 100% with settable precision. A callback function is called with the computed progress. 
I use it with long simulations to display the simulation progress.

I guess the conflict is with the committed PR #223 which adds ps simulation resolution. 
This PR also allows ps resolution plus adds the use of other configuration parameters that are no more usable since @block introduction.

NicoPy:
I have generated new PRs.
#252 for simulation configuration
#253 for progress callback

PR #207, opened 2016-12-13
Adding __reversed__ method

### LINKED ISSUE #195: reversed() doesn't work on intbv
I declared an intbv and I could iterate through the bits in normal order. But if I used reversed() on the intbv to iterate through the bits in the opposite order, I got the same result and no error message.

```
In [1]: from myhdl import *

In [2]: v = intbv('01010001')

In [3]: for b in v:
   ...:     print(b)
   ...:
False
True
False
True
False
False
False
True

In [4]: for b in reversed(v):
   ...:     print(b)
   ...:
False
True
False
True
False
False
False
True

```

tokencolour:
Trying to fix #195 
Implementation of __reversed__ method which goes opposite to the direction __iter__ method chooses.

### DIFF
--- a/myhdl/_intbv.py
+++ b/myhdl/_intbv.py
@@ -111,6 +111,12 @@ class intbv(object):
             raise TypeError("Cannot iterate over unsized intbv")
         return iter([self[i] for i in range(self._nrbits - 1, -1, -1)])
 
+    # reversed method
+    def __reversed__(self):
+        if not self._nrbits:
+           raise TypeError("Cannot reverse unsized intbv")
+        return iter([(self[i]) for i in range(self._nrbits)])
+
     # logical testing
     def __bool__(self):
         return bool(self._val)
--- a/myhdl/_intbv.py
+++ b/myhdl/_intbv.py
@@ -115,7 +115,7 @@ class intbv(object):
     def __reversed__(self):
         if not self._nrbits:
            raise TypeError("Cannot reverse unsized intbv")
-        return iter([(self[i]) for i in range(self._nrbits)])
+        return iter([self[i] for i in range(self._nrbits)])
 
     # logical testing
     def __bool__(self):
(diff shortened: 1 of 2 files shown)


tokencolour:
I've tried to implement the reversed method as discussed [here](https://github.com/myhdl/myhdl/issues/195). I understand that it is still pending to decide what route to take if an intbv is reverse iterated. I feel that when an intbv is iterated, the MSB should be taken first (this is intuitive (to me) and perhaps not very pythonic). Hence I have implemented `__reversed__` method to take element 0 first. @jandecaluwe , @cfelton , @josyb, Can you please look into this and tell if this is the right course?

tokencolour:
About the tests:
I have faced an anomaly and I can't figure out why this is happening. 
The added `__reversed__` method yields expected results when checked from the top directory (which holds the git clone).
screenshot from 2016-12-13 17 58 25
but when checked from within top/myhdl/test/core, it fails. 
screenshot from 2016-12-13 18 01 37
I can't understand why this is happening. Please help me find out what I am missing.

josyb:
I feel very strongly that the actual iterator uses the wrong direction.
 ``` 
for b in vector:
    print(b)
```
 should give the same result as 
```
for i in range(len(vector):
   print(vector[i])
```

**Comment by josyb** [comment]
Perhaps this _test_ is more convincing?
```
for i, b in enumerate(vector):
    assert b == vector[i]`
```

tokencolour:
The test is certainly convincing, if we decide to change the order `__iter__` pursues. I'll try to change my tests along the lines of the above test. Should I just change `__iter__` and `__reversed__`  methods and commit?

josyb:
@tokencolour we will have to wait for our BDFL to decide ...

tokencolour:
@josyb  I'll wait. Thanks!

jck:
>  but when checked from within top/myhdl/test/core, it fails.

This is because the version of MyHDL you are editing is not the one you have installed in your environment.

tokencolour:
@jck Thanks for the heads-up. It was indeed the same issue you said it was. All the tests pass now. Though, Travis complains about [inconsistent tabs](https://travis-ci.org/myhdl/myhdl/jobs/183567533). Could you please help me there? I used the normal 4 spaces tabs. Also, what do you think about the order of `__iter__` (and `__reversed__`)?

Issue #171, opened 2016-05-27
Bad verilog conversion


mkatsimpris:
The following code:

const = [-2, -3]

@block
def negative_assignment(a, b):

    #bug
    s1 = Signal(intbv(const[0], min=-10, max=10))
    s2 = Signal(intbv(const[1], min=-10, max=10))

Produces bad verilog code:

assign s2 = 5'd-3;
assign s1 = 5'd-2;

sergyatkeysight:
Sorry Im really new with github.  It seems this issue still has not been resolved?  What needs to be done to fix it.  I see the code to fix it was already provided by qrqiuren (tested and it works on my end).

josyb:
The PR was closed by qrqiuren ...

PR #216, opened 2017-03-31
Add test for issue #171

alvihabib:
`~$ py.test --sim=iverilog test_issue_171.py`

```
...
$ issue_171.v:22: syntax error
$ issue_171.v:22: error: syntax error in continuous assignment
$ issue_171.v:23: syntax error
$ issue_171.v:23: error: syntax error in continuous assignment
$ Analysis failed
```

PR #217, opened 2017-04-03
Solve Issue #171

qrqiuren:
The issue #171 has been solved by detecting the sign of the initial value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ work_vcom/
 # Test artifacts
 myhdl/**/*.v
 myhdl/**/*.vhd
+
+# PyCharm
+.idea/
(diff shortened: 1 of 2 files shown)


hgomersall:
Is this an issue with VHDL, or is it just verilog?

hgomersall:
This really needs an accompanying test as well...

qrqiuren:
@hgomersall Another person has added a test in #216. But I will consider a test later.

It is just for Verilog at this moment. I will take a look to the corresponding code.

hgomersall:
You shouldn't consider a test later, you should write the test first - or at least include the test with the commit. To be clear, the test is a top level component of the implementation.

Issue #222, opened 2017-05-02
VCD trace files only support 1ns timescale.


nfarring:
It appears that the current code to generate VCD trace files has hardcoded the value '1ns' as the timescale. Would be good to also support '1ps' timescales.

I have added a feature to allow the user to specify the timescale in the _Block.config_sim() call, e.g.

`inst.config_sim(trace=True, timescale='1ps')`

Will generate a pull request for this issue.

cfelton:
@nfarring, the 0.9 api supported user defined timescale and the new 0.10 api should as well ... at some point someone had a PR/proposal to add the rest of the sim options to the `config_sim` , it should work like the convert function.

As a work around you can `myhdl.traceSignals.timescale = '1ps'`.

Thanks for offering to generate the pull-request, we will keep our eyes out for it.

nfarring:
@cfelton, thank you for pointing out the workaround!

PR #223, opened 2017-05-02
Added support for picosecond resolution in VCD trace files.

nfarring:
- Added optional timescale parameter to _Block.config_sim()
  - Default value is '1ns'
- Forward timescale parameter to traceSignals() call.
- Added code to _traceSignals.py to use optional timescale keyword arg.
- Checked that both '1ns' and '1ps' show up correctly in VCD trace file.

### DIFF
--- a/myhdl/_block.py
+++ b/myhdl/_block.py
@@ -275,10 +275,10 @@ class _Block(object):
             setattr(converter, k, v)
         return converter(self)
 
-    def config_sim(self, trace=False):
+    def config_sim(self, trace=False, timescale='1ns'):
         self._config_sim['trace'] = trace
         if trace:
-            myhdl.traceSignals(self)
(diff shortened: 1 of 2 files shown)


nfarring:
Resolves issue #222 .

myhdl/_block.py
```diff
@@ -275,10 +275,10 @@ def convert(self, hdl='Verilog', **kwargs):
             setattr(converter, k, v)
         return converter(self)
 
-    def config_sim(self, trace=False):
+    def config_sim(self, trace=False, timescale='1ns'):
```
cfelton: This is ok for now but we really need a solution for all of the trace and simulation options.  This probably needs to be extended - made generic to pass a bunch of args (kwards) and then the function needs to set the sim and trace parameters.

NicoPy: This is already done in PR #201.

PR #227, opened 2017-05-23
Changed backup vcd filename creation

hellow554:
When there is already a vcd file with the same name, do not just append the timestamp to the filename, but remove the '.vcd' extension, append the timestamp and append the .vcd extension again, so that file will always have the right file extension (very useful for windows systems)


qrqiuren:
I think the corresponding test case should also be changed. Please read the log of the Travis CI.

punkkeks:
Any reason why this isnt merged yet?

qrqiuren:
@jandecaluwe

PR #231, opened 2017-06-27
Add test for neg_assignment issue

alvihabib:
Add test for neg_assignment issue using '@block'


qrqiuren:
Actually I also have tried to solve this. Mentors gave me some suggestions: #217

PR #234, opened 2017-08-08
Fixed a typo in the VHDL conversion of concat

cpeppenster:
The signals should be wrapped inside an unsigned() instead of an unsigned'().


cpeppenster:
Guess I've learned something new today, please regard this pull request as not requested ;)

PR #239, opened 2017-11-12

wsnyder:
This fixes issue 238.


wsnyder:
Cancel this, will separate the commits.

PR #240, opened 2017-11-12

wsnyder:


wsnyder:
Pull request 243 succeeds this.

PR #241, opened 2017-11-12
Compile warnings, comments

wsnyder:
This pull requests makes some trivial improvements to the cosim tree. This includes fixing a gcc warning, cleaning up some ignored files, removing blank lines at the end, and adding comments to help other people writing VPI code.

This does not change simulation results.

Thanks for your consideration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ work/
 work_nvc/
 work_vlog/
 work_vcom/
+obj_dir/
 *.wlf
 
 # Test artifacts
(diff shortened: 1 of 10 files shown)


josyb:
Unfortunately this is also 2 separate changes in a single commit; you need to create separate branches  (local in your myhdl fork) for each pull request.

wsnyder:
Actually one was on a separate branch, looks like github merges master changes with branch changes that base on that master. Will redo.

PR #243, opened 2017-11-12

wsnyder:
This fixes bug 238. Hopefully by itself this time.... cherry picked a single commit.


wsnyder:
Sorry, lost track of this and I'm not looking at myhdl integration any longer.

PR #244, opened 2017-11-12
For icarus cosim, fix some gcc warnings and add some comments. No cha…

wsnyder:
…nge to simulations.


wsnyder:
Sorry, lost track of this and I'm not looking at myhdl integration any longer.

PR #250, opened 2018-02-28
VHDL convertion : Remove unsigned() cast

NicoPy:
Remove unsigned cast for signed values when converting to VHDL.

I need this patch in my designs since I use signed intbv() a lot.
Without this patch, my designs don't work.

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -1502,9 +1502,9 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
             self.write(post)
             return
         pre, suf = self.inferCast(node.vhd, node.vhdOri)
-        if isinstance(node.value.vhd, vhd_signed) and isinstance(node.ctx, ast.Load):
-            pre = pre + "unsigned("
-            suf = ")" + suf
+#        if isinstance(node.value.vhd, vhd_signed) and isinstance(node.ctx, ast.Load):
+#            pre = pre + "unsigned("
+#            suf = ")" + suf
         self.write(pre)
         self.visit(node.value)
         lower, upper = node.slice.lower, node.slice.upper


josyb:
Unfortunately GHDL fails quite a few tests, which indicates to me that you tried to solve a *corner case* and got away with you initial code change. (It has happened to me also ...)
Perhaps you can supply a small example, or better a test module, to show your intention?

NicoPy:
I've noticed the failure but kept it like this to let others argument on the modification/current situation.
I don't understand this part of the code. 
I am sure this part of the code can't generate correct VHDL with signed signals. Forcing a signed signal to be cast to unsigned can't be correct.

josyb:
Still, show some example that generates incorrect VHDL code.

The MyHDL doc specifies that the slice of an intbv is always *posiitive* (and treated as such by the simulation). Hence the *unsigned* cast.
I understand your concern especially when the slice includes the sign bit.

NicoPy:
I know, I have to provide an example. I just don't have time for now.

NicoPy:
Here is an example :

```python
from myhdl import block, Signal, intbv, always_seq

def AudioSample(reset_value=0, nb_bits=24):
    return intbv(0, min=-(2**(nb_bits-1)), max=2**(nb_bits-1))      # min included ; max excluded

@block
def truncate(i_Clk, i_Sample, o_Sample):
    norm_size = len(i_Sample) - len(o_Sample)
    @always_seq(i_Clk.posedge, reset=None)
    def compute_proc() :
        o_Sample.next = i_Sample[:norm_size] + 1
    return compute_proc
            
def Convert():
    i_Clk    = Signal(bool(0))
    i_Sample = Signal(AudioSample(0, 30))
    o_Sample = Signal(AudioSample(0, 24))
    dut = truncate(i_Clk, i_Sample, o_Sample)
    dut.convert(hdl='VHDL', name="truncate", path=".", initial_values=True)
Convert()
```
i_Sample and o_Sample are signed bit vectors.

The problem is with the line 
```o_Sample.next = i_Sample[:norm_size] + 1```
It converts to 
```o_Sample <= signed(unsigned(i_Sample(30-1 downto 6)) + 1);```

The addition is done with an unsigned bit vector while is should be done with a signed bit vector.

**Comment by josyb** [comment]
a workaround:
```python
        o_Sample.next = (i_Sample >> norm_size) + 1
```
[this what a programmer would write :)]
produces
```VHDL
        o_Sample <= resize(shift_right(i_Sample, 6) + 1, 24);
```
The synthesis tool might just optimize it to a slice operation ...

jandecaluwe:
This PR fails regressions.

NicoPy:
@josyb 
```o_Sample.next = (i_Sample >> norm_size) + 1``` does not exactly perform the same operation. The addition is done with the size of the i_Sample signal. In my case I want to minimize the resources used. The synthesizer might optimise the adder... or not.

Re-reading one of your posts :

> The MyHDL doc specifies that the slice of an intbv is always positive (and treated as such by the simulation). Hence the unsigned cast.

You are right. I forgot it.
Thinking about it again, one obvious solution to my problem is 
```o_Sample.next = i_Sample[:norm_size].signed() + 1```

We can close this PR :)

PR #251, opened 2018-02-28
Add unsigned() method to _intbv class

NicoPy:
Adds unsigned() method to cast a signed intbv signal to an unsigned intbv signal.

### DIFF
--- a/myhdl/_intbv.py
+++ b/myhdl/_intbv.py
@@ -547,3 +547,30 @@ class intbv(object):
             return intbv(retVal, min=-M, max=M)
         else:
             return intbv(retVal)
+
+    def unsigned(self):
+        ''' Return new intbv with the values interpreted as unsigned
+        '''
+
+        # value is considered unsigned
+        if self.min is not None and self.min < 0 and self._nrbits:
+
+            # get 2's complement value of bits
+            msb = self._nrbits - 1
+
+            sign = ((self._val >> msb) & 0x1) > 0
+
+            # mask off the bits msb-1:lsb, they are always positive
+            mask = (1 << msb) - 1
+            retVal = self._val & mask
+            # if sign bit is set, subtract the value of the sign bit
+            if sign:
+                retVal -= 1 << msb
+
+        else:  # value is returned just as is
+            retVal = self._val
+
+        if self._nrbits:
+            return intbv(retVal)[self._nrbits:]
+        else:
+            return intbv(retVal)        


josyb:
I have this in my code base:
```python
    # 27-07-2014 jb
    def unsigned(self):
        '''
        in some cases an intbv holding a signed value (i.e. with min < 0) has to be
        transported over an 'Unsigned' bus e.g. the Avalon MM datapath
        '''
        if self.min < 0 and self._nrbits > 0:
            # it has a negative minimum
            msb = self._nrbits - 1
            sign = ((self._val >> msb) & 0x1) > 0
            if sign:
                # actually negative ..
                # remember that self._val is negative ...
                retVal = 2 ** self._nrbits + self._val
            else:
                retVal = self._val  # is positive
        else:
            # it is an unsigned
            retVal = self._val

        return retVal
```
which is slightly simpler. It doesn't have all sanity checks, though.
I wrote this at the time but I have only used it once I believe, for the purpose in the docstring.

cfelton:
I am trying to recall why `unsigned` was not proposed at the same time as `signed` was added.  The gmane mailing-list doesn't appear to be accessible any more (all those conversations lost forever).  

Since I don't recall any argument against an `unsigned` function this is a reasonable addition.

I have converted in the past by masking the signed `unsigned_signal.next = signed_signal & 0xFFFFF` but  you have to build the correct mask (as the example unsigned implementation), the `unsigned` function keeps it DRY.

cfelton:
@josyb are you requesting @NicoPy to adopt your conversion but keep the checks?

cfelton:
Since we are calculating an `int` (the `_val`), can't the conversion simply be made by a simple multiply?

```python
retval = -1 * int(self._val)
```

(the `int(...)` is redundant but kept as a check and an indication of what is expected).

josyb:
@cfelton 
the * - 1 is a bit naive :)
Try it with a 4- bit intbv() and you will see.
No I didn't want to impose my code, but I think it shows the two's complement paradigm better

My code could be simplified to:
```python
    # 27-07-2014 jb
    def unsigned(self):
        '''
        in some cases an intbv holding a signed value (i.e. with min < 0) has to be
        transported over an 'Unsigned' bus e.g. the Avalon MM datapath
        '''
        if self.min < 0 and self._nrbits > 0:
            # it has a negative minimum
            if  self._val < 0:
                # actually negative ..
                # remember that self._val is negative ...
                retVal = 2 ** self._nrbits + self._val
            else:
                retVal = self._val  # is positive
        else:
            # it is an unsigned
            retVal = self._val

        return retVal
```
or even more (but getting a bit cryptic also?):
```python
    # 27-07-2014 jb
    def unsigned(self):
        '''
        in some cases an intbv holding a signed value (i.e. with min < 0) has to be
        transported over an 'Unsigned' bus e.g. the Avalon MM datapath
        '''
        if self._nrbits > 0:
            # it has a negative minimum
            return self._val + 2 ** self._nrbits if self._val < 0 else 0
        else:
            # it is an unsigned
            return self._val 
```

NicoPy:
I've written the unsigned() function by copying the signed() function and modifying it where needed.

cfelton:
@josyb yes of course, I seem to have a knack for making stupid comments in the github threads (my mind was forcing abs at that point in time).

I like the conversion in your comment, but it should return an `intbv` same as `signed` - no?

josyb:
Nicolas,

My apologies.
That signed() function was too difficult for me :)
That can also be done differently.
```python
  . . .

        if self.min is not None and self.min >= 0 and self._nrbits:
            if self._val & (1 << (self._nrbits - 1)):
                # sign bit is set
                # get 2's complement value of bits
                retVal  = self._val - 2 **self._nrbits
            else:
                # it is positive so we don't need to complement it
                retVal  = self._val

        else:  # value is returned just as is
            retVal = self._val
  . . .

```
or shorter (and again a bit more cryptic and taking some effort to decode) :
```python
  retval = self._val - (2 ** self._nrbits if self._val & (1 << (self._nrbits - 1)) else 0)
```
The parenthesis around the subtrahend are quite necessary, so I have to correct the unsigned() also:
```python
    return self._val + (2 ** self._nrbits if self._val < 0 else 0)
```

josyb:
@cfelton 
github threads are forgiving ;)
The intbv() encapsulation is not necessary as ```.signed()``` and ```.unsigned()``` are always at the RHS of an assignment?
But I guess adding it might be safer; as I said I only used it once and maybe even not.

cfelton:
@josyb only to be consistent with `signed`, in my opinion they both should return `int` or `intbv`, but not one return `intbv` (signed) and one `int` (unsigned).

josyb:
@cfelton 
Totally agree. Consistency is key.

cfelton:
@josyb where do we sit on this PR?  Is it good to go?  Do you want changes, if so we should call out specifically in the diff/review section.  

We have to finalize which PRs are merged for the 0.10 release.

josyb:
@cfelton 
I am OK with @NicoPy's original input. I just couldn't resist being my pedantic self :)

PR #253, opened 2018-02-28
Adds simulation progress callback management.

NicoPy:
Adds the possibility to pass a user callback function to simulation run.
The user callback is called when the progress percentage changes.
The progress percentage precision is user definable.

### DIFF
--- a/myhdl/_Simulation.py
+++ b/myhdl/_Simulation.py
@@ -110,7 +110,7 @@ class Simulation(object):
     def quit(self):
         self._finalize()
 
-    def run(self, duration=None, quiet=0):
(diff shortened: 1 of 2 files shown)


josyb:
The callback only works if you also specify a ```duration``` which limits this PR to a small number of simulations as most users will run simulation with ```duration=0``` and raise a ```StopSimulation``` when done.
I'm also convinced that you can accomplish this functionality outside the MyHDL package by e.g. extending the clock generation process.

NicoPy:
@josyb you are right, this works only when duration is not 0.

NicoPy:
This PR works only with a non null duration. 
Also, I agree its job can be done outside MyHDL.
Closing it.

PR #254, opened 2018-02-28
VHDL conversion : Add variable initialisation

NicoPy:


josyb:
GHDL fails 6 tests ...
Your code imposes initialisation for every variable.  Can you explain, with an example, why you want this?

NicoPy:
Looks like I have to modify the tests.

Variables can have default values exactly like signals do.
Also, constant values can be translated to VHDL variables :
```python
        # My constant values
        PREAMBLE_X = intbv("11100010")[8:] 
        PREAMBLE_Y = intbv("11100100")[8:] 
        PREAMBLE_Z = intbv("11101000")[8:]

        @always_seq(i_Clk.posedge, None)
        def proc_tx():
            if i_Preamble == PREAMBLES.X :
                tx_preamble.next = PREAMBLE_X[7:0]
            elif i_Preamble == PREAMBLES.Y :
                tx_preamble.next = PREAMBLE_Y[7:0]
            elif i_Preamble == PREAMBLES.Z :
                tx_preamble.next = PREAMBLE_Z[7:0]
```
Here, PREAMBLE_X, PREAMBLE_ and PREAMBLE_Z are converted to variables.
As variables are not initialised, the converted design does not work.

josyb:
To me they look to be *constants*? (as you say in the comments ...)
So you can and should rewrite your code rather than modify MyHDL.

NicoPy:
My code is correct. It simulates correctly and it converts correctly except the variable initialisation. 
It took me hours to find the bug in the converted code.
Either my code should not convert, either this is a converter bug.

How do you suggest to rewrite my code ?
Why do you think variables should not be initialised ?

NicoPy:
I fixed the bug which made Travis CI  tests fail.

josyb:
A variable serves a purpose: usually to break a long assignment into smaller steps.
The conversion should have picked up the fact that the *variable* is not written to and have issued a warning.
You are using it as a constant, so a constant is more appropriate.
You could write your code as something like:
```python
    # My constant values
    PREAMBLE_X = int("1100010", base=2)
    PREAMBLE_Y = int("1100100", base=2)
    PREAMBLE_Z = int("1101000", base=2)

    @always_seq(i_Clk.posedge, None)
    def proc_tx():
        if i_Preamble == PREAMBLES.X :
            tx_preamble.next = PREAMBLE_X
        elif i_Preamble == PREAMBLES.Y :
            tx_preamble.next = PREAMBLE_Y
        elif i_Preamble == PREAMBLES.Z :
            tx_preamble.next = PREAMBLE_Z

```
I removed the leading'1' in each constant to make it work ...
Note that the names PREAMBLE_X and fellows will not show up in the converted code, but you will see numbers instead.
Alternatively you wrap the constant in a Signal:
```python
    # My constant values
    PREAMBLE_X = Signal(intbv("11100010")[8:])
    PREAMBLE_Y = Signal(intbv("11100100")[8:])
    PREAMBLE_Z = Signal(intbv("11101000")[8:])

    @always_seq(i_Clk.posedge, None)
    def proc_tx():
        if i_Preamble == PREAMBLES.X :
            tx_preamble.next = PREAMBLE_X[7:0]
        elif i_Preamble == PREAMBLES.Y :
            tx_preamble.next = PREAMBLE_Y[7:0]
        elif i_Preamble == PREAMBLES.Z :
            tx_preamble.next = PREAMBLE_Z[7:0]

```
I am not sure about this one though. It will probably generate a signal, and issue a warning about it. Perhaps we need a ```Constant``` class or perhaps a ```constant=True``` argument to ```Signal```? Although you can set the ```read```attribute to True to silence the warning (in MyHDL only, the vendors' VHDL compilers will still issue a warning, but that warning will go lost in the flood) 
 If you had provided a small example code I would have tried it first. I didn't immediately find a candidate in my code-base to test this.

Actually in VHDL we can also specify a constant in a process where we declare variables; that would be the best approach (IMHO). But it may be quite different in Verilog (which we shouldn't neglect)

NicoPy:
I've tried the Signal() solution. I had to set .driven attribute to True to remove warnings.

My intent is not to use variables in this case. At first, I thought the converter will generate constants. But that is not the case ; variables are generated.

Your solution using int() is not correct from my point of view since the the initialisation binary value has to be modified knowing the use case. And it would not be valid if there were another use of the constant with another slicing.

josyb:
You specify a variable, you get a variable.  
Actually MyHDL doesn't generate VHDL constant declarations at all.

I only see a very small window of your code, so I can only work with that. I knew you were going to rebut me on the int() approach. I'm sure if I can see the whole code I can make an elegant work-around.

NicoPy:
> You specify a variable, you get a variable.

Yes, and No. There is no way to define a constant. The syntax I used is the most straight forward way of writing.

However, I think we are out of context here.
I'll send you a private message on discourse with a real use case.

The point here is : Do we want to add variables initialisation to converted VHDL code ?
(I believe the same is possible with Verilog but I don't use verilog.)

josyb:
> Yes, and No. There is no way to define a constant. The syntax I used is the most straight forward way of writing.

Yes, MyHDL doesn't do constants (yet?)

> The point here is : Do we want to add variables initialisation to converted VHDL code ?

Yes, we should, although we can live without it, as we can initialise the variable as the first action in the process.

The point is that you actually want a constant ... using a variable to achieve this is not pure (and thus not beautiful IMHO). So we may start thinking of adding some way of specifying a constant in MyHDL. (for the next release, after 0.10)

NicoPy:
> So we may start thinking of adding some way of specifying a constant in MyHDL. (for the next release, after 0.10)

 👍

NicoPy:
I've found that this patch generates wrong VHDL in many situations. 
To avoid this problem, variables must be defined before use outside of the process (like in VHDL). This is not acceptable.
It can be very complex to fix this issue, so it's better to reject this PR.

A "Constant" type is very welcome

cfelton:
@NicoPy did you set `inital_value` when converting?

@NicoPy I agree this PR should be closed for now.  The initial value support has slowly been added, originally some (Verilog) tools did not support, and we can't manage one IR supporting features the other doesn't.  The idea of variable initial value, needs some additional thought and consideration as well as the Verilog implementation.

NicoPy:
@cfelton  Yes, ```initial_values``` has been set. The problem is deeper. Globally, most initialisations are correct. But in some conditions (that I don't know), the initial value written in the VHDL file is "something". One workaround is to declare the variable outside the process function and use this variable inside the function.

@cfelton  I agree that the verilog part should be written. But not by me. My verilog knowledge is too small for this. 
I have not closed this PR since it is a base for working on this functionality if we want it to be implemented after version 0.10.

josyb:
@NicoPy 
In the mean time I have added a `Constant` declaration (subclassed from `Signal`).
Is oit OK to close this PR?

Reagrds,
Josy

PR #255, opened 2018-03-01
Reports range() without argument correctly

NicoPy:
When using range() without parameter, a cryptic error message is displayed by converter :
```
File "C:\Python34\lib\site-packages\myhdl-1.0.dev0-py3.4.egg\myhdl\conversion\_toVHDL.py", line 1275, in visit_For
    start, stop, step = args
ValueError: need more than 0 values to unpack
```

Replaces it by a useful information message :
```
myhdl.ConversionError: in file xxx.py, line 49:
    Requirement violation: at least one argument requested
```

### DIFF
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -1208,6 +1208,7 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
         args = cf.args
         assert len(args) <= 3
         self.require(node, len(args) < 3, "explicit step not supported")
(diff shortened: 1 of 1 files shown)


josyb:
The no-argument case would normally have flagged by the simulator?

NicoPy:
Sure.
Sometimes you quickly write a small module, don't test it, use it. I know, this is bad practice, this is real life.

josyb:
The rule should be to simulate always ... or live with the consequences.
But certainly the more useful error message is a good idea.

NicoPy:
As we say in French : "Ceinture et bretelles" ;)

PR #256, opened 2018-03-10
A fix for initial value memory conversion and user code

cfelton:
If a list-of-signals is referenced in user defined code,
the conversion fails.

The string of the list-of-signals (LOS) was written instead
of the name of the reference to the LOS.  A function was
added to resolve the name instead of the raw string
conversion of the type.

### DIFF
--- a/myhdl/_extractHierarchy.py
+++ b/myhdl/_extractHierarchy.py
@@ -115,7 +115,14 @@ class _UserCode(object):
         code = "\n%s\n" % code
         return code
 
+    def _scrub_namespace(self):
+        for nm, obj in self.namespace.items():
+            if _isMem(obj):
+                memi = _getMemInfo(obj)
+                self.namespace[nm] = memi.name
+
     def _interpolate(self):
+        self._scrub_namespace()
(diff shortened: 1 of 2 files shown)


cfelton:
@hgomersall @josyb please review this PR, I think this needs to be included in the 0.10 release.

cfelton:
I need to add the VHDL user code version ...

hgomersall:
Looks fine to me. Is there a reason why there is a second level of test block? It seems to me to be a slightly complicated test scenario (reading in from a hex file).

cfelton:
@hgomersall the hex file thing is needed for the `$readmemh` function, and that is it (not really used).  I needed the two-levels for the user-code, otherwise the signals are not declared if I user-code the top-level function.

hgomersall:
I that case, I have no concerns. Do I need a proper review?

cfelton:
I removed the WIP, this PR still lacks a VHDL test case but I will not have time to add the test case in the next week or so.  It is safe to merge and the VHDL test case can be added in the future.

PR #257, opened 2018-03-16
intbv.unsigned() internal enhancement

NicoPy:
While the current unsigned() method of intbv() class gives correct results, its internal coding is not optimal.
This is a patch to make internal code more efficient and more "conceptually correct".


cfelton:
@NicoPy is there a way to reopen / append to the previous "merged" unsigned PR?

NicoPy:
@cfelton  If there is one, I don't know it.

NicoPy:
I have a patch to make unsigned convertible but only for VHDL.

PR #260, opened 2018-03-24
EnumType is allowed as free variables

hashhsah:
EnumType is integer-like, and should be allowed where integer is allowed, I think.

### DIFF
--- a/myhdl/conversion/_analyze.py
+++ b/myhdl/conversion/_analyze.py
@@ -626,7 +626,7 @@ class _AnalyzeVisitor(ast.NodeVisitor, _ConversionMixin):
             if f.__code__.co_freevars:
                 for n, c in zip(f.__code__.co_freevars, f.__closure__):
                     obj = c.cell_contents
-                    if not isinstance(obj, (integer_types, _Signal)):
+                    if not isinstance(obj, (integer_types, _Signal, EnumType)):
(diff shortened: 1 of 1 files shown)


cfelton:
@hashhsah can you provide an example before and after.  A case where it failed, and an case where it is allowed now?  Also, is there any tests that can go with this?

hashhsah:
@cfelton it was long time ago, let me dig out my old code

hashhsah:
@cfelton Below please find a minimal example. I honestly can't remember the usage case I originally intended.

`c` is a *free variable* to the function `my_mode()`. Since enum is mapped to integer during conversion, it should be allowed whenever int is allowed.

```python
from myhdl import always_comb, Signal, intbv, instances, block, enum

MODE = enum('MODEA', 'MODEB')

@block
def mod1():
    A = Signal(False)
    B = Signal(False)
    S = Signal(intbv(0)[32:0])
    Y = Signal(False)

    c = MODE.MODEA

    def my_mode():
        #return c               # will fail
        return MODE.MODEA       # converts fine

    @always_comb
    def comb():
        if S==my_mode():
            Y.next = A
        else:
            Y.next = B
    return instances()

mod1().convert()

```

I just realize the error message for `FreeVarTypeError` should probably be revised too:

```python
class _error(object):
    ...
    FreeVarTypeError = "Free variable should be a Signal or an int"
    ...
```

PR #264, opened 2018-07-06
Renamed async variables to remove namespace conflict with new keyword introduced in Python V3.5

BluDotz:
In and from Python V3.5, the keyword "async" was introduced to make asynchronous programming a fully first-class element of the language.

This caused a name conflict with similarly-named member variables in the MyHdl module, so those have been replaced with the non-conflicting "isAsync".

The MyHdl module now loads without error in Python V3.7.

### DIFF
--- a/myhdl/_always_seq.py
+++ b/myhdl/_always_seq.py
@@ -45,7 +45,7 @@ _error.EmbeddedFunction = "embedded functions in always_seq function not support
 
 class ResetSignal(_Signal):
 
-    def __init__(self, val, active, async):
+    def __init__(self, val, active, isAsync):
         """ Construct a ResetSignal.
 
         This is to be used in conjunction with the always_seq decorator,
@@ -53,7 +53,7 @@ class ResetSignal(_Signal):
         """
         _Signal.__init__(self, bool(val))
         self.active = bool(active)
-        self.async = async
+        self.isAsync = isAsync
 
 
 def always_seq(edge, reset):
@@ -91,8 +91,8 @@ class _AlwaysSeq(_Always):
         if reset is not None:
             self.genfunc = self.genfunc_reset
             active = self.reset.active
-            async = self.reset.async
-            if async:
+            isAsync = self.reset.isAsync
+            if isAsync:
                 if active:
                     senslist.append(reset.posedge)
                 else:
--- a/myhdl/conversion/_toVHDL.py
+++ b/myhdl/conversion/_toVHDL.py
@@ -1838,12 +1838,12 @@ class _ConvertAlwaysSeqVisitor(_ConvertVisitor):
         senslist = self.tree.senslist
         edge = senslist[0]
         reset = self.tree.reset
-        async = reset is not None and reset.async
+        isAsync = reset is not None and reset.isAsync
         sigregs = self.tree.sigregs
         varregs = self.tree.varregs
         self.write("%s: process (" % self.tree.name)
         self.write(edge.sig)
-        if async:
+        if isAsync:
             self.write(', ')
             self.write(reset)
         self.write(") is")
@@ -1853,7 +1853,7 @@ class _ConvertAlwaysSeqVisitor(_ConvertVisitor):
         self.writeline()
         self.write("begin")
         self.indent()
-        if not async:
+        if not isAsync:
             self.writeline()
             self.write("if %s then" % edge._toVHDL())
             self.indent()
@@ -1870,7 +1870,7 @@ class _ConvertAlwaysSeqVisitor(_ConvertVisitor):
                 self.write("%s := %s;" % (n, _convertInitVal(reg, init)))
             self.dedent()
             self.writeline()
-            if async:
+            if isAsync:
                 self.write("elsif %s then" % edge._toVHDL())
             else:
                 self.write("else")
@@ -1881,7 +1881,7 @@ class _ConvertAlwaysSeqVisitor(_ConvertVisitor):
             self.writeline()
             self.write("end if;")
             self.dedent()
-        if not async:
+        if not isAsync:
             self.writeline()
             self.write("end if;")
             self.dedent()


alexforencich:
This needs to be merged ASAP.  Arch Linux just updated to python 3.7, now myhdl is unusable.

alexforencich:
To expedite this, I'll make an updated patch that also updates all of the tests.  What's the best replacement?  isAsync?  isasync?  Something else?

BluDotz:
Hi Alex

I used “isAsync” in the patch I submitted via the pull request. I thought it gave a useful hint that the variable was a boolean.

But of course it’s your module, so please decide what you feel is best.

Cheers,
Philip

alexforencich:
It's not my module, I just use myhdl extensively and have a keen interest in un-breaking it.

BluDotz:
Apologies. I thought you were the module owner.

How do we twist the module owner’s arm to make the changes and update the release?

alexforencich:
Not sure.  Looks like Jan has been AWOL since April.  I think fixing the PR to get the tests to pass would expedite the process of getting this merged.  I presume making that happen is just a matter of replacing any instances of async in the tests.  Now, as for what to replace it with, it should be consistent with the rest of the code, and the rest of the code seems to avoid camelCase in arguments, hence why I suggest isasync instead of isAsync, but perhaps there is something else that would be more concise.

BluDotz:
If one “follows” him on GitHub, does that allow one to PM him?

BluDotz:
Roger that. I’ve sent him an email.

cfelton:
@alexforencich and @BluDotz thanks for raising this issue and submitting a PR.  I will help facilitate getting this issues resolved ASAP.  However, I am currently swamped with other things as well.  We are working on some options to fix this flow when those with commit rights (@jandecaluwe, @jck) become overwhelmed with other responsibilities.  

As @alexforencich pointed out, because some of the linux distros updated to 3.7 (wow, bleeding edge), this  should be fixed as soon as possible.

I would call it `isasync` or move to `sync`, @josyb or @hgomersall  opinion the argument name?

josyb:
Of course I have a different name in mind: why not use its full name ```asynchronous```? Every modern IDE (Eclipse-PyDev, PyCharm, ... ) has auto-completion so we should no longer be afraid of using long names.
I would like to stay away from ```sync``` and thus *inverting* the logic; I often use ```Reset = ResetSignal(0, 1, True)```

alexforencich:
If we can agree on the best replacement for 'async', I can go ahead replace all of the relevant instances of 'async' in the current master and submit an updated PR.

BluDotz:
Ah, now I see why no changes ever get committed; everyone’s spending waaaayyyyy too much time yak shaving (https://en.wiktionary.org/wiki/yak_shaving <https://en.wiktionary.org/wiki/yak_shaving>). FFS choose some vaguely suitable names already.

alexforencich:
Hah, I haven't heard of yak shaving before, I'll have to remember that one

alexforencich:
Replaced 'async' with 'asynchronous' in new PR here: https://github.com/myhdl/myhdl/pull/267 .

BluDotz:
Whether it’s yak shaving or not, the effect is the same: no commit means no-one who uses Python V3.5 or higher can use myhdl. Time to pull your finger out.

cfelton:
@BluDotz @hgomersall , I agree.  I think it is ridiculous to have little patience to discuss the naming of public API functions.   If this was some internal name that no one else had to deal with - yes *yak shaving* probably appropriate.

We have bigger issues of not being able to respond timely to issues like this.

BluDotz:
Taking the pedantic view, yes we are changing the names used in a public API but, the significant point is, we are not changing the method signature in any way. Otherwise we wouldn’t be wasting time debating what the method variable name is.

It is not as if this is some deliberative change which surely does require deep thought; it is nothing less than a necessary response to a change in the Python language reserved words.

alexforencich:
Sheesh.  I submitted an updated PR here, with 'async' swapped out for 'asynchronous' in what I believe is every instance in the repo, aside from the release notes.  It passes all tests save for the GHDL tests as it seems GHDL has broken itself again.  So all we need to do is fix the GHDL test environment (more yak shaving, unfortunately) and we should be good to go.

hgomersall:
Stop with the fscking Yak shaving. You're doing nobody any favours by belitting opinions - you're just wasting yet more time.

The _real_ issue here, is getting stuff merged. Perhaps you can think about how to deal with that problem, because at the moment that's what's stalling development.

BluDotz:
If I was in a position to do anything about it I would. But I’m not, so I can’t. I’m stuck - like everyone else - sitting on my hands until someone with the authority makes the appropriate commit.

My PR dates back to 6th July, now well over a month ago. And the actual issue which the PR offered to resolve goes as far back as Python 3.5, which was released on 13th September 2015 (https://docs.python.org/3/whatsnew/3.5.html <https://docs.python.org/3/whatsnew/3.5.html>) !

I have every sympathy with the current committers if they are over-stretched on other projects, but then either a) they need to authorise additional committers for this project or b) pass it over to someone for whom it is an essential part of their daily workflow. In both cases @alexforencich <https://github.com/alexforencich> would seem to be an ideal candidate.

hgomersall:
@BluDotz the problem, as I've no doubt you've realised, not one either me or @cfelton or @josyb can fix either. We too a frustrated by this issue.

josyb:
>If I was in a position to do anything about it I would. But I’m not, so I can’t. I’m stuck - like everyone else - sitting on my hands until someone with the authority makes the appropriate commit.

You can temporarily revert to Python 3.6

BluDotz:
The relevant change to the Python language happened on and from V3.5 ((https://docs.python.org/3/whatsnew/3.5.html <https://docs.python.org/3/whatsnew/3.5.html>), so the current version of the myhdl module will fail on any and all versions of Python since then.

One would actually need to revert to Python 3.4 - the version before the new async keyword was introduced.

hgomersall:
@BluDotz Not reserved until 3.7: https://docs.python.org/3/whatsnew/3.7.html

BluDotz:
Ok. Now hanging my head in shame :-(

Mucho apologies to @josyb .

cfelton:
Addressed in #267

PR #267, opened 2018-08-11
Rename async to isasync to resolve conflict with python 3.7

alexforencich:
Python 3.7 makes 'async' a syntax error.  This renames all instances of 'async' to 'isasync' to fix this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,8 @@ env:
   - CI_TARGET=iverilog
   - CI_TARGET=ghdl
 
-# Remove when py37 bugs are fixed
 matrix:
   allow_failures:
-    - python: "3.7"
     - python: "pypy3.5"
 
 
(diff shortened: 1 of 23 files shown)


alexforencich:
Hmm, it looks like almost all of the GHDL tests failed.  Does this change really break all of that, or did something in the test environment change?

josyb:
It is (yet again) a different path (or a different method all together?) to get GHDL working in Travis-CI
We will have to call on @jck  to help us out.

alexforencich:
Updated PR to replace 'async' with 'isasync' as per gitter discussion

jck:
@alexforencich please rebase to pick up travis changes, and create a new commit removing python 3.7 as an allowed failure(delete line 23).

Alternatively, you can "Allow edits from maintainers" on the right sidebar.

alexforencich:
Looks like the GHDL tests passed under 3.6, but GHDL failed to install with python 3.7??

alexforencich:
Ah, it seems it failed to retrieve a PGP key.  Very strange.  Maybe that one just needs to be restarted?

cfelton:
passed after a restart.

PR #270, opened 2018-08-24
Nested Top-Level Interfaces

josyb:
Added conversion of nested interfaces for the top-level module/function

### DIFF
--- a/example/manual/tb_inc.v
+++ b/example/manual/tb_inc.v
@@ -1,4 +1,4 @@
-module tb_inc;
+module tb_Inc;
 
 wire [7:0] count;
 reg enable;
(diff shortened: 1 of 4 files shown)


josyb:
Actually there is an issue converting to VHDL when using *std_logic_Ports=True*. The *xxx_yyy_num* signals will not be created.

hackfin:
> My opinion is that we should reconsider that second test. I remember a discussion quite a while ago where I claimed that that _renaming_ better resulted in an error.

I'd second that. The current state is (see also #286) that the top level interface is again 'broken'. It looks like there should be a VHDL specific test case scenario? (Couldn't find one yet)

PR #277, opened 2018-10-08
enum: added string output for enums in _traceSignals.py

josyb:
Currently MyHDL encodes enum as real in VCD output. Some VCD viewers, e.g. Impulse by toem.de, don't support this well but have a true string representation. As has GTKWave.
This PR rplaces the *real* by a *str* encoding.

### DIFF
--- a/myhdl/_traceSignals.py
+++ b/myhdl/_traceSignals.py
@@ -23,7 +23,6 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
-
 import sys
 import time
 import os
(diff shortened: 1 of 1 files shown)


josyb:
@cfelton @hgomersall @jck @jandecaluwe 
Would you be so kind to review this PR?

myhdl/_traceSignals.py
```diff
@@ -212,11 +213,19 @@ def _writeVcdSigs(f, hierarchy, tracelists):
                 siglist.append(s)
             w = s._nrbits
             # use real for enum strings
-            if w and not isinstance(sval, EnumItemType):
-                if w == 1:
-                    print("$var reg 1 %s %s $end" % (s._code, n), file=f)
+            if w:
```
jck: can't this entire block be simplified to something like this?
```
            # use string for enum strings
            if w and not isinstance(sval, EnumItemType):
                print("$var reg %s %s %s $end" % (w, s._code, n), file=f)
            else:
                print("$var string 1 %s %s $end" % (s._code, n), file=f)
```

PR #290, opened 2018-12-16
Enhanced ShadowSignals

josyb:
Added the possibility to take a ShadowSignal of a whole Signal, which is useful if you e.g. want to build several different ListOfSignals from a pool of Signals.
Split the SliceSignal() class over 3 new classes: CloneSignal, SliceSignal and IndexSignal to simplify the structure and improve maintainability (and readability)
Added a *signed* keyword to allow cutting up a large incoming vector into signed objects.
Expanded the allowed string types in ConcatSignal.

### DIFF
--- a/myhdl/test/conversion/general/test_ShadowSignal2.py
+++ b/myhdl/test/conversion/general/test_ShadowSignal2.py
@@ -39,7 +39,7 @@ def bench_SliceSignalSigned():
         print(h)
         aa.next = e + f
         # iverilog and ghdl seem to need a delta delay
-        yield delay(0)
+        yield delay(1)
         print(aa)
 
         a.next = 127


josyb:
Still needs adding additional tests!

hashhsah:
very interesting feature. kudos

I'm tempted to abuse this to implement memory-mapped registers with bit fields. I'm curious what the intended application is.

josyb:
You can get rid of ShadowSignals by replacing the ```intbv``` by a more pure ```BitVector```. Until then they remain a necessity; as you once explained me several years ago ...

hgomersall:
I'm being stupid, but I don't understand this. Can we have a quick justification?

josyb:
@hgomersall doesn't the top description not say what it does?

PR #297, opened 2019-02-03
update installation instructions

### LINKED ISSUE #295: Update installation instructions in readme
* add note recommending to use virtual environments
* change examples to use pip instead of `python setup.py install`
* add instruction for installing directly from git: `pip install -e 'git+https://github.com/myhdl/myhdl#egg=myhdl'`

jck:
fixes #295

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,28 +38,30 @@ To find out what's new in this release, please read:
 
 Installation
 ------------
-If you have superuser power, you can install MyHDL as follows:
+It is recommended to install MyHDL (and your project's other dependencies) in
+a virtualenv.
+
+Installing the latest stable release:
 
 ```
-python setup.py install
+pip install myhdl
 ```
 
-This will install the package in the appropriate site-wide Python
-package location.
-
-Otherwise, you can install it in a personal directory, e.g. as
-follows:
-
+To install the development version from github:
 ```
-python setup.py install --home=$HOME
+pip install -e 'git+https://github.com/myhdl/myhdl#egg=myhdl
 ```
 
-In this case, be sure to add the appropriate install dir to the
-``$PYTHONPATH``.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,28 +38,30 @@ To find out what's new in this release, please read:
...
 ```
-python setup.py install --home=$HOME
+pip install -e path/to/dir
 ```
 
-In this case, be sure to add the appropriate install dir to the
-``$PYTHONPATH``.
+To install a specific commit hash, tag or branch from git:
+```
+pip install -e 'git+https://github.com/myhdl/myhdl@f696b82ca5c9e3668920d1b66263941dcbcf08b6#egg=myhdl
```
cfelton: For a specific hash can you use a 6-digit shorthand?

README.md
```diff
@@ -38,28 +38,30 @@ To find out what's new in this release, please read:
 
 Installation
 ------------
-If you have superuser power, you can install MyHDL as follows:
+It is recommended to install MyHDL (and your project's other dependencies) in
+a virtualenv.
```
cfelton: Should include a link to virtualenv.

```[virtualenv](https://packaging.python.org/guides/installing-using-pip-and-virtualenv/)```
[virtualenv](https://packaging.python.org/guides/installing-using-pip-and-virtualenv/)

README.md
jck: Yeah, anything that works with `git checkout`

README.md
jck: how would you feel about linking to this instead? https://docs.python-guide.org/dev/virtualenvs/

It recommends `pipenv`, which in my opinion is currently the best tool for managing a project's dependencies, but also explains the other options such as directly creating a virtualenv or using virtualenvwrapper.

PR #307, opened 2019-03-12
Drop support for Python 2

### LINKED ISSUE #301: Pledge to drop Python 2 support
https://python3statement.org/

Many major projects including pytest, ipython, pandas and numpy are dropping Python 2 support.

We can simplify some of our code, and explore Python 3 only features such as asyncio by dropping Python 2 support in a later version.

We should decide on a timeline for the first Python 3 only version.

hugovk:
Fixes #301.


hugovk:
Updated to resolve conflicts.

jck:
This is great. Thanks!

PR #315, opened 2019-05-12
Updated docs for 0.11 release

cfelton:
This PR mainly includes document updates for the 0.11 release.  But there are some other changes in the examples/manual modules, because I originally thought some of these were causing the RTD build to fail.  Instead of reverting them I left them.


cfelton:
I need to revert the README changes, docs should point to the original location.

PR #319, opened 2019-08-01
Conversion : Enable list in ports.

NicoPy:
VHDL only.
Verilog does not support arrays in ports.

### DIFF
--- a/myhdl/_block.py
+++ b/myhdl/_block.py
@@ -233,6 +233,7 @@ class _Block(object):
             self.vhdl_code = _UserVhdlInstance(deco.vhdl_instance, self.symdict, func.__name__,
                                                func, srcfile, srcline)
         self._config_sim = {'trace': False}
+        self.hdl = None
 
     def _verifySubs(self):
         for inst in self.subs:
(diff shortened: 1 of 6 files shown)


josyb:
VHDL-only won't do ...
Perhaps we have to  switch to generating SystemVerilog output in stead of Verilog.

NicoPy:
@josyb I don't follow you. 
The problem is, the work can't be done for Verilog. 
Why should be not publish an enhancement for VHDL because of this ?
There are already differences between VHDL and Verilog.

josyb:
> The problem is, the work can't be done for Verilog.
> Why should be not publish an enhancement for VHDL because of this ?
> There are already differences between VHDL and Verilog.

You could have worked out a solution for both Verilog and VHDL, no? By enumerating the port-array in the entity/module declaration:
```
 entity port_with_array is
  port (
   ...
   array_port_0 : in std_logic_vector ( .... );
   array_port_1 : in ....
```
I must check on what differences there are between VHDL and Verilog conversion; perhaps you can give me some pointers/examples?
Isn't it the rule that everything should convert *and* to Verilog *and* to VHDL?

NicoPy:
Test benches are convertible to Verilog but not to VHDL.

josyb:
> Test benches are convertible to Verilog but not to VHDL.

Really? I have never tried to ...

josyb:
This enhancement has been addressed for both Verilog and VHDL in PR #362 
I agree that your version is better VHDL; perhaps we should consider merging the two ideas.

PR #328, opened 2020-01-20
block decorator parameterization and simpler signal names when needed

jvillar:
Since the introduction of the block decorator, all functions in the calltrace of a design, have to be decorated by it. This way, all these functions make a contribution to the name of its internal signals. This behaviour prevents name duplication and is the desired behaviour in most cases.

But there are times that intermediate functions don’t generate hardware generators nor any kind of myhdl objects, signals, etc… However they introduce a very high complexity in the converted code, that renders it unreadable.  After some discussion at MyHDL discourse forum (https://discourse.myhdl.org/t/simplified-signal-naming-when-possible/403), I modified the MyHDL block decorator to accept  conversion arguments so the conversion process can be parameterized. In this PR a new feature is introduced to make it possible to explicitely mute some hierarchy levels regarding to signal naming. 

I think this can be a good starting point for parameterizing the conversion process.

### DIFF
--- a/myhdl/_block.py
+++ b/myhdl/_block.py
@@ -146,10 +146,12 @@ class _bound_function_wrapper(object):
         return _Block(self.bound_func, self, name, self.srcfile,
                       self.srcline, self.keepname, *args, **kwargs)
 
-class hardblock(object):
+class block_decorator(object):
 
-    def __init__(self, func, keepname=True):
-        self.keepname = keepname
+    keepname = True
+    ident_method = "get_instance_ident"
+    
+    def __init__(self, func):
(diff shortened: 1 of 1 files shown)


josyb:
Hi Jose,

I am revisiting this PR as I find a use for this in my upcoming work.
I obtained merging rights (only) in 2022.
Since then I have introduced a sub-minor version number - we currently are at 0.11.50
I searched on adding this to your PR but it looks complicated, especially as your fork has gotten quite behind.
Are you still interested? If not, can I use your code in a new PR; I will give you  due credit both in the source as in the commit message.

Best regards,
Josy

PR #329, opened 2020-02-16
Add +: verilog operator

jdavidberger:
The included test generates these error messages:

```
chunk_buffer_sim.v:36: error: A reference to a wire or reg (`chunk_buffer0_idx') is not allowed in a constant expression.
chunk_buffer_sim.v:36: error: Part select expressions must be constant.
chunk_buffer_sim.v:36:      : This lsb expression violates the rule: chunk_buffer0_idx
chunk_buffer_sim.v:36: error: A reference to a wire or reg (`chunk_buffer0_idx') is not allowed in a constant expression.
chunk_buffer_sim.v:36: error: Part select expressions must be constant.
chunk_buffer_sim.v:36:      : This msb expression violates the rule: ((chunk_buffer0_idx)+('sd8))-('sd1)
chunk_buffer_sim.v:36: error: A reference to a wire or reg (`chunk_buffer0_idx') is not allowed in a constant expression.
chunk_buffer_sim.v:36: error: Part select expressions must be constant.
chunk_buffer_sim.v:36:      : This lsb expression violates the rule: chunk_buffer0_idx
chunk_buffer_sim.v:36: error: A reference to a wire or reg (`chunk_buffer0_idx') is not allowed in a constant expression.
chunk_buffer_sim.v:36: error: Part select expressions must be constant.
chunk_buffer_sim.v:36:      : This msb expression violates the rule: ((chunk_buffer0_idx)+('sd8))-('sd1)
```

before the follow-up fix patch. 

Without that patch; it outputs this verilog:

```
        Output <= Input[(idx + 8)-1:idx];
```

and after it outputs this:

```
        Output <= Input[idx+:8];
```

### DIFF
--- a/myhdl/conversion/_toVerilog.py
+++ b/myhdl/conversion/_toVerilog.py
@@ -1216,6 +1216,15 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
         # special shortcut case for [:] slice
         if lower is None and upper is None:
             return
+
+        if isinstance(lower, ast.BinOp) and isinstance(lower.left, ast.Name) and isinstance(upper, ast.Name) and upper.id == lower.left.id and isinstance(lower.op, ast.Add):
+            self.write("[")
+            self.visit(upper)
+            self.write("+:")
+            self.visit(lower.right)
+            self.write("]")
+            return
+        
         self.write("[")
         if lower is None:
             self.write("%s" % node.obj._nrbits)
--- a/myhdl/test/conversion/general/test_colon_plus_operator.py
+++ b/myhdl/test/conversion/general/test_colon_plus_operator.py
@@ -1,0 +1,57 @@
+from myhdl import *
+
+@block
+def chunk_buffer(Clk, Reset, Input, Output):
+    OUTPUT_BIT_COUNT = len(Output)
+    idx = Signal(modbv(0)[(len(Input) // OUTPUT_BIT_COUNT):])
+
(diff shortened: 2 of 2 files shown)


hgomersall:
What happens with VHDL code? Do we have any idea how that is inferred?

hgomersall:
Otherwise, it looks good to me.

jdavidberger:
Before and after this PR, it generates this for VHDL:

```
        Output <= Input(to_integer(idx + 8)-1 downto to_integer(idx));
```

which `ghdl` seems to be fine with according to CI.

hgomersall:
I was actually pondering more about the tools inferring its a bus.

jdavidberger:
I'm not really sure; its possible it depends on the tool but this PR doesn't really modify that behavior one way or the other. I'm not familiar enough with VHDL to know whether or not there is an equivalent syntax form in it.

cfelton:
@jdavidberger I assume after almost two decades the *indexed vector part select* should be supported by all tools.  Did you happen to test with any other tools, specifically any synthesis tools?  My only worry would be that some synthesis tools might not support.  But, as I write this, I guess the users isn't any worse off.

jdavidberger:
I've run it with Synplify Pro as part of an icecube2 project. As expected, the before version errors out and the +: version works. 

That is the only one I've tried (other than iverilog), but like you hinted at, there are almost certainly no tools that would accept the prior form that wouldn't accept the indexed vector part select.

hgomersall:
If it's valid verilog then we should use it!

PR #333, opened 2020-04-10
travis: start testing on Python 3.8 too

jck:


cfelton:
@jck any reason not to merge?

PR #338, opened 2020-04-13
Added cosimulation with verilator v4 and https://github.com/csail-csg/pyverilator

joshuisken:
Cosimulation with verilator.

I used ‘pyverilator’ to ease integration. It is a specific version https://github.com/csail-csg/pyverilator which allows easy dictionary access to io-ports.
No interprocess communication is used since the verilated design is imported as a ctypes module.

Several tests have been added. Some are skipped automatically whether pyverilator is installed or not. So it should not break anything.

Automatic merge fails on myhdl/_block.py after 10-april upstream update: something easy to fix.

What I noticed is that both this verilator cosim, but also other cosims like icarus, have some initialization/reset issues which deserve some attention. They behave slightly different than pure myhdl.

Also, I considered to not use pyverilator and rewrite directly for verilator directly, but with this pyverilator version it was more convenient. but anyway, something to reconsider.

### DIFF
--- a/myhdl/_Cosimulation.py
+++ b/myhdl/_Cosimulation.py
@@ -45,6 +45,7 @@ _error.OSError = "OSError"
 class Cosimulation(object):
 
     """ Cosimulation class. """
+    name = 'cosimulation_dut'
 
     def __init__(self, exe="", **kwargs):
         """ Construct a cosimulation object. """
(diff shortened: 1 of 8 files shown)


joshuisken:
Please remove the myhdl._compat lines in both myhdl/_Verilation.py and myhdl/test/core/testVerilation.py, that should fix it.
How can I do that myself?

hgomersall:
@joshuisken Sorry, I don't understand the question!

cfelton:
@joshuisken you can change your current clone / branch and push, the pull-request will automatically be updated.

PR #339, opened 2020-04-19
Exercising "test_tristate.py"

josyb:
In an attempt to fix the second test in myhdl\test\conversion\toVerilog\test_tristate.py.
I start this PR off with a *white space change* to verify the current state passes.
Although this looks like I am abusing Travis/CI (which in some way I am as i don't get cosimulation to work under  Windows 10)  to document the process along the way.

### DIFF
--- a/myhdl/test/conversion/toVerilog/test_tristate.py
+++ b/myhdl/test/conversion/toVerilog/test_tristate.py
@@ -36,12 +36,12 @@ def tristate_obuf_i(obuf):
     # Caveat: A local name of the interface signals must be declared,
     #         Otherwise, _HierExtr.extract() will not add them to symdict
     #         and conversion will fail.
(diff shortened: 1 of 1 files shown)


josyb:
oops, bad start ...
I switched to the master branch, then created this branch and checked it out.
But apparently some changes of the last active branch are lingering around ...

josyb:
Good, we have confirmed the current state.
Let's do some a few *innocent* changes

josyb:
This works fine. (To my surprise).
Going through the commits in the original PR  #336 I see that I *adventurously* (and willfully) violated the **Caveat** in `def tristate_obuf_i(obuf):`
So I will now switch back to the other branch.

josyb:
The cause of the symptom has been remedied in #336.

PR #342, opened 2020-05-31
Update _toVerilog.py

TongJoe:
no the attribute signed in line 1604 originally.


cfelton:
@TongJoe do you have an example where this fails?  If so, can  you also provide a test that can be included in the regression tests.

TongJoe:
@cfelton ok.
    I tried the project, https://github.com/andrecp/myhdl_simple_uart, in order to simulate UART with myhdl. It was good when I run test_bench() of  tb_serial.py, but  I always got an exception when I change toVHDL() to toVerilog(). 
    Thanks for  your checking

TongJoe:
@cfelton  I find that the PR was not updated to the master. I cannot  use the _toVerilog() to convert my design to Verilog code. Can you have some help for me ?

PR #347, opened 2020-08-31
Use first value instead of -1 as index in for loops

benzea:
In some circumstances, -1 might end up being used as e.g. an intbv
index which is then invalid. To avoid this, it makes sense to stay
within the given range during analysis. Try to do so by simply picking
the first value from the range.

----

So, for the heck of it, I was trying to port some old code (myhdl 0.8, python2) to myhdl 0.10, python 3).

The line https://github.com/benzea/nco/blob/python3/prng/rule30.py#L50 fails though, because myhdl hardcodes the index to -1, and a -1 index access is invalid for intbv. A commit like this one should fix the issue, by trying to pick a value that is within the user given range.

### DIFF
--- a/myhdl/conversion/_analyze.py
+++ b/myhdl/conversion/_analyze.py
@@ -707,13 +707,18 @@ class _AnalyzeVisitor(ast.NodeVisitor, _ConversionMixin):
         self.refStack.push()
         self.visit(node.target)
         var = node.target.id
-        self.tree.vardict[var] = int(-1)
 
         cf = node.iter
         self.visit(cf)
         self.require(node, isinstance(cf, ast.Call), "Expected (down)range call")
         f = self.getObj(cf.func)
         self.require(node, f in (range, downrange), "Expected (down)range call")
+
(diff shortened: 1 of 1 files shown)


benzea:
OK, pretty sure this is wrong. A workaround is to do:

```
            a = rot_l[i]
            b = state[i]
            c = rot_r[i]
            addr = concat(a, b, c)
```

which avoids the indices inside the `concat` parameters.

hgomersall:
Can you fettle a test case that highlights the problem as well as the fix?

benzea:
Hmm, the example is not very large, i.e. the block in
  https://github.com/benzea/nco/blob/python3/prng/rule30.py#L50
has 35 LOC (a tiny sub-module of the numerical oscillator).

Unfortunately, looking into it a bit, the `concat` to build the `rot_l` and `rot_r` variables is not working anymore either. I'll probably need some more time to look into it, but this has a really low priority for me; I needed that NCO for my diploma thesis and just looked into porting things to python3 for fun one day ...

Issue #348, opened 2020-11-23
traceSignals incorrectly writes real value signal to VCD


zlobriy:
traceSignals function incorrectly writes real value signal to VCD. Verilog Standard states that a real value signal in VCD dump should start with "r" or "R" identifier. But instead, MyHDL creates signal starting with "s". So GTKwave can't make an analog wave. A function _printVcdStr caused this behavior. Just change 's' to 'r'.

NicoPy:
Can you provide a simple excerpt of code showing the problem ?

zlobriy:
Here is a simple model of ideal Digital to Analog Converter (DAC):

`from myhdl import modbv, Signal, delay, always, Simulation, traceSignals, instance, now

import numpy as np

def clock_source(clock, frequency):
    period = 1 / frequency / 10 ** (-9)  # ns
    half_period = delay(round(period / 2))

    @always(half_period)
    def clksrc():
        clock.next = not clock

    return clksrc

def digital_sin_source(clock, sine, frequency, bits):
    out_range = 2 ** (bits - 1)

    @always(clock.posedge)
    def sinsrc():
        sine.next = int(out_range * np.sin(2 * np.pi * frequency * now() / 10 ** 9))

    return sinsrc


def ideal_dac(clock, digital, analog, bits, amin=-1, amax=1):
    step = (amax - amin) / 2 ** bits

    @always(clock.posedge)
    def dac():
        analog.next = amin + step * digital

    return dac


def test_ideal_dac():
    bits = 8
    clock = Signal(bool(0))
    digital = Signal(modbv(0, min=-(2 ** (bits - 1)), max=2 ** (bits - 1)))
    analog = Signal(float(0))

    clksrc = clock_source(clock, 100 * 10 ** 6)
    sine = digital_sin_source(clock, digital, 10 * 10 ** 3, bits=bits)
    dac = ideal_dac(clock, digital, analog, bits=bits)

    @instance
    def test():
        yield delay(10)

    return clksrc, sine, dac, test

traceSignals.filename = 'test'
test_inst = traceSignals(test_ideal_dac)
Simulation(test_inst).run(100000)
`

This code generates test.vcd file. Then I open it in GTKwave (version 3.3.100) and add all signals. The "analog" signal is a real value signal, so I choose in GTKwave: Data Format -> Analog -> Interpolated. But nothing happens. When I locally modified _printVcdStr function and install MyHDL using pip from local files all is ok.

NicoPy:
[...]

PR #351, opened 2020-12-04
Fix for issue #348 (traceSignals incorrectly writes real value signal to VCD)

NicoPy:
Add a fix to correctly format real number signals in VCD files.


NicoPy:
All checks passed except the Python 3.9 ones.

PR #355, opened 2021-01-25
replace yield_fixture decorator by fixture

hellow554:
https://docs.pytest.org/en/latest/yieldfixture.html

> Since pytest-3.0, fixtures using the normal fixture decorator can use a yield statement to provide fixture values and execute teardown code, exactly like yield_fixture in previous versions.
> 
> Marking functions as yield_fixture is still supported, but deprecated and should not be used in new code.

### DIFF
--- a/myhdl/test/core/test_traceSignals.py
+++ b/myhdl/test/core/test_traceSignals.py
@@ -114,7 +114,7 @@ def topTristate():
     return inst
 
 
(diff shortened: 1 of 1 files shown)


jck:
Thanks!

Merging despite failing CI since this change is pretty safe and unrelated to the root cause of test failures

Issue #356, opened 2021-02-19
One-hot state machine conversion to Verilog


kc64:
If a state machine is implemented with IntEnum one-hot encoding for state definition, the conversion results in code like the following:
`...
else begin
	casez (io1_trigger_state)
		3'b??1: begin
			if ((io1_trigger_source && io1_gate_source)) begin
				trigger <= 1'b1;
				io1_trigger_state <= 3'b010;
			end
		end
		3'b?1?: begin
			trigger <= 1'b0;
			io1_trigger_state <= 3'b100;
		end
		3'b1??: begin
			if ((!io1_trigger_source)) begin
				io1_trigger_state <= 3'b001;
			end
		end
		default: begin
			io1_trigger_state <= 3'b001;
		end
	endcase
end
...`

This results in less-deterministic behavior since errant state vectors branch to the state that most closely matches the errant state vector and the default case, which is supposed to handle errant vectors deterministically, is never involved.

PR #357, opened 2021-02-21
Removed 'casez' in Verilog onehot / onecold state encoding

josyb:
Fixes: One-hot state machine conversion to Verilog #356

### DIFF
--- a/myhdl/conversion/_toVerilog.py
+++ b/myhdl/conversion/_toVerilog.py
@@ -471,6 +471,7 @@ def _getSignString(s):
     else:
         return ''
 
+
 def _intRepr(n, radix=''):
     # write size for large integers (beyond 32 bits signed)
     # with some safety margin
@@ -478,17 +479,18 @@ def _intRepr(n, radix=''):
     p = abs(n)
     size = ''
     num = str(p).rstrip('L')
-    if radix == "hex" or p >= 2**30:
+    if radix == "hex" or p >= 2 ** 30:
         radix = "'h"
         num = hex(p)[2:].rstrip('L')
-    if p >= 2**30:
-        size = int(math.ceil(math.log(p+1,2))) + 1  # sign bit!
+    if p >= 2 ** 30:
+        size = int(math.ceil(math.log(p + 1, 2))) + 1  # sign bit!
 #            if not radix:
 #                radix = "'d"
     r = "%s%s%s" % (size, radix, num)
-    if n < 0: # add brackets and sign on negative numbers
+    if n < 0:  # add brackets and sign on negative numbers
         r = "(-%s)" % r
     return r
+
 
 def _convertGens(genlist, vfile):
     blockBuf = StringIO()
@@ -1005,11 +1007,7 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
 #        if node.isFullCase:
 #            self.write(" full_case")
 #        self.writeline()
-        caseType = "case"
-        if isinstance(node.caseItem, EnumItemType):
-            if node.caseItem._type._encoding in ('one_hot', 'one_cold'):
-                caseType = "casez"
-        self.write("%s (" % caseType)
+        self.write("case (")
         self.visit(var)
         self.write(")")
         self.indent()
@@ -1017,7 +1015,7 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
             self.writeline()
             item = test.case[1]
             if isinstance(item, EnumItemType):
-                self.write(item._toVerilog(dontcare=True))
+                self.write(item._toVerilog())
             else:
                 self.write(self.IntRepr(item, radix='hex'))
             self.write(": begin")
@@ -1215,7 +1213,7 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
             self.visit(lower.right)
             self.write("]")
             return
-        
+
         self.write("[")
         if lower is None:
             self.write("%s" % node.obj._nrbits)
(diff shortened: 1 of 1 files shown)


hgomersall:
Presumably this just removes the special case of one-hot and one-cold?

josyb:
Yes, it only applies to `enum( xxx, yyy, ..., encoding='one_xxxx')`
see also: https://discourse.myhdl.org/t/one-hot-state-machine-encoding-in-verilog-conversion/452

hgomersall:
So my understanding of the discourse discussion is that somehow having "don't cares" is bad - what's the rationale here? I suspect that explicit full values would be optimised to the same synthesized output, but given the conversion is from myhdl, I can't see the strong argument for changing this behaviour...

josyb:
The converted Verilog code is no good, because it infers a priority encoder. If the state machine is well behaved it will work fine, but the case of detecting multiple bits goes lost, so the `default` case  has no effect.

hgomersall:
Is the concern that MyHDL will generate a misbehaving state machine, or that there is redundant code in a `default`? Arguably in a converted enum there should be no default.

It's not clear to me how the two types are optimised - Is a priority encoder always single path? Can it be optimised better than fully elaborated branches?

josyb:
A one_hot or one_cold state machine is mainly chosen for safety to recover illegal states and to recover; the converted Verilog code currently voids   this `default` statement.
A priority encoder is 'cheaper' to implement than the fully decoded case. Although the vendor tool probably will recognize the one-hot/one-cold state machine and act on each single bit, unless multiple bits are set in which case it will take the default route.

hgomersall:
But why would an illegal state occur? Is this about dodgy code or cosmic rays flipping bits?

josyb:
cosmic rays is one of the kind

hgomersall:
Fine, in which case this PR seems good.

JanCoombs:
I am concerned that there appears to be false assumption and presumption in the logic leading to this change, and that this has lead to three significant problems involving speed, complexity, and inappropriate behaviour. 
Or maybe just one problem - no closed-loop testing?
Is there any sample verilog output available for a simple state machine from before and after this change?  It would be quicker for me to run tests than to do a write-up.

josyb:
See #356 
The changes in the produced Verilog code will be:
before:
```python
	casez (io1_trigger_state)
		3'b??1: begin
			if ((io1_trigger_source && io1_gate_source)) begin
				trigger <= 1'b1;
				io1_trigger_state <= 3'b010;
			end
		end
		3'b?1?: begin
			trigger <= 1'b0;
			io1_trigger_state <= 3'b100;
		end
		3'b1??: begin
			if ((!io1_trigger_source)) begin
				io1_trigger_state <= 3'b001;
			end
		end
		default: begin
			io1_trigger_state <= 3'b001;
		end
	endcase
```
after:
```python
	case (io1_trigger_state)
		3'b001: begin
			if ((io1_trigger_source && io1_gate_source)) begin
				trigger <= 1'b1;
				io1_trigger_state <= 3'b010;
			end
		end
		3'b010: begin
			trigger <= 1'b0;
			io1_trigger_state <= 3'b100;
		end
		3'b100: begin
			if ((!io1_trigger_source)) begin
				io1_trigger_state <= 3'b001;
			end
		end
		default: begin
			io1_trigger_state <= 3'b001;
		end
	endcase
```

JanCoombs:
Thanks Josy, now I only need to find half a free day.

JanCoombs:
beste Josy, 

Thanks for the sample code. I am about two days into intensive testing, 
and so far looks as expected. Code size and speed vary with the number 
signals input to the state machine, and with size. Have tested with 4 
and 8 states, then noticed this odd problem with 16 states:

Is there a requirement for the enum identifiers to all be of the same 
length? The conversion to Verilog has an extra 'default: begin case' 
inserted between states 'STTs.S9:' and 'STTs.S10'. This goes away if 
the identifier names are all of equal length, as in 'STTs.S09'.

Hope all is going well with you. We are also moving soon, so there will 
be much shortage of time, and painting to be done. 

Best wishes to you, Jan.
--

josyb:
Hi Jan,

The length of the names should have no influence at all. But there will always be a default case, which preferably is put at the end. Perhaps in the 16-way state machine it got inserted between 9 and 10 and not at the end. Or did you get 2 default statements?

Don't overdo yourself with moving, it is a stressful and fatiguing period.

Regards,
Josy

JanCoombs:
Tests of two state machines with 4, 8, or 16 states did not produce expected results. This was partly due to poor managment of testing, and likely also that the tests were too contrived. 

To test with a real example I coded the Phase Lock Loop state machine, as documented in the HEF4046 PLL chip data. After setting up a crude [digital] VCO model, it was possible to remove typos, and then see it locking onto an incoming signal. Although unstable it produce an exact edge-count match to the simulated incoming signal. 

This PLL state machine has 12 states, each mostly having two source states, and two target states, both conditional.  Summary results are:

1hot+'?'s   111 MHz  237 CLBs
1hot+'0's   120 MHz  208 CLBs
binary       130 MHz  182 CLBs
  CLBs = LUT4+FF     MHz = fMax

For this small example, using an open-source tool chain[1], the current change to one_hot encoding improves performance of the Verilog export. However binary encoding is better still.

I would like to understand what the synth tool is actually doing, so this is going on my bucket list. The tools have an option to convert the bitstream back into Verilog, but this is a  very low-level Verilog representation, and offers only a small step towards recognising the synthesized code. 

So all of my complaint and ill speculation is withdrawn for now, and likely much longer. 
[1] Project IceStorm   [http://bygone.clairexen.net/icestorm/](http://bygone.clairexen.net/icestorm/)

PR #359, opened 2021-03-14
Conversion with recursive ShadowSignals

josyb:
When using recursion and taking ShadowSignal from a ShadowSignal the conversion (both Verilog and VHDL) looses the connection with the starting Signal. This PR fixes that.

_ShadowSignal.py: conversion:
  enabled taking a ShadowSignal of a ShadowSignal,
  reworked _setName() to aggregate chained slicing/indexing to a single slice or index
  marked the shadowed Signal as  always 'read'
_intbv.py: added the [n:] to __repr__
_toVHDL.py: added a comment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ myhdl/**/*.vhd
 
 # Pycharm ide junk
 .idea/
+/.pytest_cache/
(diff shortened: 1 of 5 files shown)


josyb:
Of course, this PR still fails for Python3.9 ...

hgomersall:
The point here is that you can chain shadow signals, right?

josyb:
> The point here is that you can chain shadow signals, right?

Yes, *chaining* might have been a better description than *recursive*

PR #362, opened 2022-02-24
Added support for lists in ports

lucasasselli:
For both VHDL and Verilog. Similar to what @josyb proposed in #319.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ work_vcom/
 # Test artifacts
 myhdl/**/*.v
 myhdl/**/*.vhd
+myhdl/**/*.hex
 
 # Pycharm ide junk
 .idea/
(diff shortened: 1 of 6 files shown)


hgomersall:
To be clear, this is top level ports?

hgomersall:
It looks good as an approach. One thing that concerns me is something that I've been worrying about for a while which is that we don't do any checking of name collisions after mangling (unless that has been fixed?).

hgomersall:
Ok, it seems the name checks are in place already, at least for VHDL. Each port name should be checked using `_nameValid`.

josyb:
I have to correct my previous post: even if we do the 'enumeration' in _analyze.py we still have some changes i _toVHDL.py and in _toVerilog.py.
But I also found out that there is more to test.

PR #371, opened 2022-06-08
[ENH] Updated the ConcatSignal conversion code to check for undriven …

tobygomersall:
…signals.

I have seen a bug when a ConcatSignal is created in the following way:

    ```sig_list = [Signal(False) for n in range(32)]
    concat_sig = ConcatSignal(*reversed(sig_list))```

It seems that the _name attribute on the signals in sig_list is only updated if an assignment is made to one of the signals in sig_list. Otherwise _name is left as None. None should not be written in to the resultant VHDL or Verilog.

This pull request add codes to check the signal names for None and raise a warning if present. If the signal name is None then the updated code treats it as a constant and drives it with the specified initial value.

### DIFF
--- a/myhdl/_ShadowSignal.py
+++ b/myhdl/_ShadowSignal.py
@@ -196,8 +196,39 @@ class ConcatSignal(_ShadowSignal):
             else:
                 w = len(a)
             lo = hi - w
+
+            if isinstance(a, _Signal) and a._name is None:
+                # We have seen a bug when a concat signal is created in the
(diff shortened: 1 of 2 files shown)


josyb:
@tobygomersall 
Can you add a test?

tobygomersall:
Yea, I'll add one.

tobygomersall:
@josyb
I have added a test. Sorry for the delay.

PR #375, opened 2022-07-31
SystemVerilog support

josyb:
A MyHDL user had an issue with dynamic signal generation where the conversion  generated `'wire` where a `reg` was expected.
> So the VHDL builds correctly and the Verilog one does not there is a bug in the Verilog version - it would be easy in Systemverilog as everything would be 'logic' and not he old 'wire/reg' thing.
As it is not the VHDL is correct - if you change to verilog it tries to make a clocked process with a wire - which is an error.

This PR :
*provides initial SystemVerilog support*:
  - changing the module header format to e.g.;
 ```verilog
module square_root (
    input  wire clock,
    input  wire clock_sreset,
    input  wire data_valid,
    input  wire [11:0] data,
    output logic  result_valid,
    output logic  [11:0] result,
    output logic  [11:0] remainder
    );
```
- replacing `wire` and `reg`  by `logic` where appropriate

*adds a docstring to config_sim()*


josyb:
Closing this PR as I am working on a new converter strategy with a separate SystemVerilog conversion module (separated from the Verilog converter module) anticipating further enhancements (Structure, (true) Array, (true) Interface, hierarchical conversion)

PR #381, opened 2022-10-23
Reduce memory footprint

telemarkguru:
Some small optimizations to reduce the memory used. Especially when generating Verilog or VHDL.
It allocates some Signal member variables lazily to avoid allocation memory not needed.


JanCoombs:
Documentation is sparse, sample project demonstrating retained functionality and improvements seems to be missing. 

Drastically reducing memory footprint' is not always a good sign. Why would each signal wait list (which I assume is) created during analysis be still available when starting the simulation?

telemarkguru:
Hi,
> Documentation is sparse, sample project demonstrating retained functionality and improvements seems to be missing.
I've updated the commit message with more details. It was indeed a bit sparse.
Do you think some sample code is needed here when no functionality is changed?
If so, where should it be added?

> 
> Drastically reducing memory footprint' is not always a good sign. Why would each signal wait list (which I assume is) created during analysis be still available when starting the simulation?

I don't quite follow you here. This change only postpones the allocation of signal wait lists, and the signal slice list to when they are first used, and once allocated there is no difference to how it worked before.  The gain comes from that all these lists (4 in total) are rarely used by most Signals. One would e.g. expect that the negedge and posedge lists are usually used by clocks and asynchronous reset Signals.

Best regards,
-stefan

josyb:
The changes to self._slicesigs are unnecessary as iterating an empty list doesn't raise an exception, and what currently happens for 99% of the Signals

telemarkguru:
On Thu, Oct 27, 2022 at 2:01 PM Josy Boelen ***@***.***>
wrote:

> The changes to self._slicesigs are unnecessary as iterating an empty list
> doesn't raise an exception, and what currently happens for 99% of the
> Signals
>
But self._slicesigs is now initialized to None to avoid allocating a list
when not used, so I think the checks are needed.
Or do you refer to something else?
Best Regards,
Stefan

josyb:
I did delete that comment, as I realized you saved the cost of creating the empty list ... and the None cannot be iterated

You can save yet a little more by doing alike for self._code

The _tradeoff_ we now have is less memory usage (and consequent speed-up) for conversion (and for the initial simulation too I guess) against more tests when running simulation e.g. the 1M*128 bit will be quick to allocate, but a simulation that exercises the full memory will now take longer because of the additional tests for None; and in the end every entry in the memory block will have been updated after all.
I guess there is more experimenting/studying to do?

josyb:
Perhaps we could create _shallow_ Signals when converting using either the combined  to `_verilog._converting` / `_toVHDL._converting` flags or introducing new _global_ to `_verilog._shallow` / `_toVHDL._shallow` flags?
I gave that a quick try but got stuck with _circular imports_ :(

telemarkguru:
> I did delete that comment, as I realized you saved the cost of creating the empty list ... and the None cannot be iterated
> 
> You can save yet a little more by doing alike for self._code
> 
> The _tradeoff_ we now have is less memory usage (and consequent speed-up) for conversion (and for the initial simulation too I guess) against more tests when running simulation e.g. the 1M*128 bit will be quick to allocate, but a simulation that exercises the full memory will now take longer because of the additional tests for None; and in the end every entry in the memory block will have been updated after all. I guess there is more experimenting/studying to do?

You're right.  I made a quick simulation performance measurement on a RAM model, and indeed it ran slightly slower.
But, when updating the code to use try/except to handle when _eventWaiters is None, the simulation performance actually improved slightly over master.  Interesting.

josyb:
That last commit was for co-simulation and not about reducing the memory footprint. So I'd rather see it in a separate branch.
Although it passes all tests (see above) the win32 part is not tested - as we still have to add that to GitHub actions

davekeeshan:
> @davekeeshan can we add a single win32 action, say for Python 3.11 only?

I think we can add the general testing, against 3.11 yes, getting the tests to run against iverilog and ghdl on windows on CI is a little more involved. The current setup is to compile it.

There is a ghdl action I have been meaning to investigate:

https://github.com/marketplace/actions/setup-ghdl

There isn't one for icarus yet

davekeeshan:
I have opened a PR, #410, that adds windows support for `core` and `ghdl` under python3.11.

I have iverilog mostly working, it gets installed from choco. (it is currently v11_0) however when the github action tries to build the vpi, it falls over:

I would apprecaite if any one has any insights into getting around [this](https://github.com/daxzio/myhdl/actions/runs/4124620230/jobs/7124099015):

```
mingw32-make[1]: Entering directory 'D:/a/myhdl/myhdl/cosimulation/icarus'
iverilog-vpi myhdl.c myhdl_table.c
x86_64-w64-mingw32-gcc.exe: fatal error: input file 'myhdl.c' is the same as output file
compilation terminated.
myhdl_table.c:3:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
    3 | extern void myhdl_register();
      | ^~~~~~
myhdl_table.c:5:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
    5 | void (*vlog_startup_routines[])() = {
      | ^~~~
iverilog-vpi: 1 file(s) failed to compile.
Compiling myhdl.c...
Compiling myhdl_table.c...
mingw32-make[1]: *** [makefile:4: myhdl.vpi] Error 2
mingw32-make[1]: Leaving directory 'D:/a/myhdl/myhdl/cosimulation/icarus'
mingw32-make: *** [makefile:54: iverilog_myhdl.vpi] Error 2
Error: Process completed with exit code 2.
```

davekeeshan:
@josyb has approved and merged #410, to pick up that change in this branch please do a merge from master into your branch and re push it, it should all rerun naturally

Issue #388, opened 2022-12-07
verliog generation of constants is strange


olisnr:
myHDL is perfekt to simultate and generate verilog for FPGAs. 
myHDL constants will be converted in the verilog code to numbers. this is not so nice, but works in most cases. But today i had problems whit one constant in a verliog if-statment. in myHDL the constant was "CLK_DiV =  4" and the if-statement "elif clkdiv == CLK_DiV // 2:". So i would expect an if-statment that compares to 2 in verilog, unfortunatly the generator produced: "(-'h1): begin" and the syntesizer optimized this out... i tried different things:
```
            elif clkdiv == CLK_DiV // 2: # gives "(-'h1): begin"
            elif clkdiv == math.ceil(CLK_DiV / 2): # gives error: "Not supported: true division - consider '//'"
            elif clkdiv == 2: # gives "'h2: begin"
            elif clkdiv == CLK_PH2: # gives "'h2: begin" this works and using "CLK_PH2 = CLK_DiV // 2"
```

## System information
```
MyHDL Version: 0.11 from "pip install git+https://github.com/myhdl/myhdl.git@master"
Python Version:  3.9
```

my full code is:

```
import math

import random
from myhdl import block, always, instance, Signal, delay, StopSimulation, intbv, instances, ResetSignal, always_comb, always_seq



RESET_ACTiVE =  0
ADC_CHiPs =     4
ADC_DATA_BiTs = 18
ADC_SCLK_CONV = 16  #500ns
CLK_DiV =       4   #max 32MHz SCLK
CLK_PH1 =       0
CLK_PH2 =       CLK_DiV // 2

CLK_DIV_WiDE = math.ceil(math.log2(CLK_DiV))
CLK_CNTR_WiDE = math.ceil(math.log2(ADC_DATA_BiTs + ADC_SCLK_CONV + 3))




@block
def qaud_adc_spi(    clk,
                        reset_n,
                        enable,
                        
                        sclk,
                        conv,
                        to_adc,
                        from_adcs, 
                        
                        data_out0, 
                        data_out1, 
                        data_out2, 
                        data_out3, 

                        data_out_new
                    ):



    #intern
    shifts =    [Signal(intbv(0)[ADC_DATA_BiTs:]) for _ in range(ADC_CHiPs)]
    # cntr =      Signal(intbv(0, max=ADC_DATA_BiTs + ADC_SCLK_CONV + 3))
    # clkdiv =    Signal(intbv(0, max=CLK_DiV))
[...]

PR #389, opened 2022-12-09
Fixes issue #388 incorrect conversion of `elif siga ==  expression`

josyb:
_toVerilog.py:
  corrected mapToCase conversion of 'test'; replacing integer output by visit(test.comparators[0])
  this will properly convert tests like e.g.  (el)if siga == constb // integer

### DIFF
--- a/myhdl/conversion/_toVerilog.py
+++ b/myhdl/conversion/_toVerilog.py
@@ -1046,7 +1046,9 @@ class _ConvertVisitor(ast.NodeVisitor, _ConversionMixin):
             if isinstance(item, EnumItemType):
                 self.write(item._toVerilog())
             else:
-                self.write(self.IntRepr(item, radix='hex'))
+                # we can assume there will only be one comparison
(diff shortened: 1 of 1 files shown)


davekeeshan (approved):
Small enough change, all I would ask if there is a test possible to should this working (not working before the change) but not worth requesting a change (which is one of the options below)

davekeeshan (approved):
Sorry github wasn't refreshing, probably updating permissions, I logged out and in to find that my review had actually occurred.

josyb:
Dave,

We could use the AO's file and make a test out of it.
I actually found a few more discrepancies: if I add an 'else: pass` to the if-elif chain the  VHDL conversion will then also map this if-elif-else to a case and produce the same type of error, but instead of (-h1) produce "11" ( -1 converted to 2 bits). If I repair this in the same manner as Verilog I get:
image
The Sigasi linting flags this as an error - I believe it shouldn't so I raised a question at Sigasi 
And then I also pondered we need more tests.

PR #398, opened 2022-12-16
Add reverse daxzio

### LINKED ISSUE #195: reversed() doesn't work on intbv

davekeeshan:
This is a pull request to remap this one, #207, that was made before the github actions were added to the test

> Trying to fix #195
> Implementation of reversed method which goes opposite to the direction iter method chooses.

Merge this one and close the other one.


josyb:
@davekeeshan 
I am hesitant to merge this PR as I have questions about the iteration order. (see my original comments in #207)
we have:
```python
    # iterator method
    def __iter__(self):
        if not self._nrbits:
            raise TypeError("Cannot iterate over unsized intbv")
        return iter([self[i] for i in range(self._nrbits - 1, -1, -1)])
```
which in my opinion is wrong because:
```python
vector = intbv(0x12)[8:]
i = 0
for b in vector:
    print(i, b, vector[i])
    i += 1
```
gives this result:
```
0 False False
1 False True
2 False False
3 True False
4 False True
5 False False
6 True False
7 False False
```
adding this `__reversed__` method will only confuse us more?
The other question is: if we correct the iteration order, how much user code will we break?

davekeeshan:
@josyb I have to admit, I am not entirely clear on what the subtlty here is.  I was using the pull request more as a demo to show how to upgrade old requests so that they could pull the the github action flow.

However the questions are valid, and entirely in your hands.  However I think you/we should decide to implement this change or close the request, having them hang around in this zombie state is messy, someone else could be having htis conversation again in another 6 years (don't forget the data contained in a close pull request is still available and can be reopened.)

Not to your actual question, first, yes I am confused. 

The original issue is that looping through intbv and looping through reversed(inbv) was giving the same result, on the surface that is just wrong.

Maybe the answer is that the code changed with this PR is wrong, and should be closed, but the issue that was rasied, which is open, is still valid and is awaiting the correct implementation?

I need to trawl through the code base to give a better answer than that (but I will if i have to to get these requests closed!)

davekeeshan:
@josyb is the issue not with reversed, more that the original iter is actually wrong? It appears be deliveing data in reversed order? and the addtio with this PR is that is is correctly reversing what iter is doing, but iter is wrong?

If so, if you fix this bug (? maybe that is actually what needs an issue to capture the problem) you break peopls code base? 

If so maybe it needs a 2 stage fix, a deprecation/legacy stage?

davekeeshan:
OK I have investigated further.  If this was an issue I think it was fixed a long time ago.  

I ran this test and it passed:

```
import warnings
def testIter():
    val = 0x12
    b = [
        False,
        True,
        False,
        False,
        True,
        False,
        False,
        False,
    ]
    vector  = Signal(intbv(val)[len(b):])

    assert vector[0] == b[0]
    assert vector[1] == b[1]
    assert vector[2] == b[2]
    assert vector[3] == b[3]
    assert vector[4] == b[4]
    assert vector[5] == b[5]
    assert vector[6] == b[6]
    assert vector[7] == b[7]

    i = 0
    for x in vector:
        warnings.warn(f"{i} {x} {vector[i]}")
        assert x == vector[i]
        if 7 == i:
            break
        i += 1

    i = 0
    for v in reversed(vector):
        warnings.warn(f"{i} {v} {b[len(b)-1-i]}")
        assert v == b[len(b)-1-i]
        if 7 == i:
            break
        i += 1
```
However I found another problem, the reason why I had to add breaks into the loops is that the values being returned for the loops never terminated.  On investigation this for loop is not populated by the `__iter__ `method but the `__getitem__`, in fact I commented out the `__iter__ `segment of code and no test failed in the test suite, I added an exception into the method and it never was rasied.

```
def testEnumerate():
    val = 0x12
    vector  = Signal(intbv(val)[8:])
    
    assert 8 == len(vector)
    
    for i, v in enumerate(vector):
        if i > 8:
            raise Exception("Too many value returned from vector")
```
`FAILED myhdl/test/conversion/general/test_print.py::testEnumerate - Exception: Too many value returned from vector`

Here the loop continues beyond the length of the intbv, so much so it never terminates a loops for ever, only a kill command rescued my computer.

What is wrong in the `__getitem__ ` method is that the `IndexError`nneds to be raise when the index used exceeds the size of the vector.


```
    def __getitem__(self, key):
        if isinstance(key, slice):
            i, j = key.start, key.stop
            if j is None:  # default
                j = 0
            j = int(j)
            if j < 0:
                raise ValueError("intbv[i:j] requires j >= 0\n"
                                 "            j == %s" % j)
            if i is None:  # default
                return intbv(self._val >> j)
            i = int(i)
            if i <= j:
                raise ValueError("intbv[i:j] requires i > j\n"
                                 "            i, j == %s, %s" % (i, j))
            res = intbv((self._val & (1 << i) - 1) >> j, _nrbits=i - j)
            return res
        else:
            i = int(key)
            res = bool((self._val >> i) & 0x1)
            return res
```
The else statement needs to be changed to:
```
        else:
            i = int(key)
            if i >= self._nrbits:
                raise IndexError('End')
            res = bool((self._val >> i) & 0x1)
            return res
```
However this works as I would expect, it seems to break exisiting tests.  

I think this is the right thing to do, but as you said earlier something like this might break exisiting designs.

davekeeshan:
Sorry some confusion and mixed up branches on my part.  

The order of iteration in the master branch is still "wrong", but the indexing issue seems to be a legit bug.  I will switch to the issues associated with those bugs, which is probably the more appropriate place to discuss them, rather than the PR

josyb:
IMO there is no indexing bug
```python
from myhdl import intbv
vector = intbv(0x12)[8:]
for b in vector:
    print(b)
```
produces the expected result (in the wrong order, of course):
```
calling on intbv:__iter__
False
False
False
True
False
False
True
False
```
and this code:
```python
i=0
for b in vector:
    print(f'{i} {b} {vector[i]}')
    i += 1
print('Done!')
```
also runs without hangs or exceptions:
```
calling on intbv:__iter__
0 False False
1 False True
2 False False
3 True False
4 False True
5 False False
6 True False
7 False False
Done!
```

davekeeshan:
@josyb Yes, my messed up branch merges don't show it the same way, and it doesn't appear as an issue on the iterator.

However when one tries to access a index  of a vector that does not exist, the __getitems__ should raise and indexError, currently it does not.

You may say that this is not an issue, but I think an error should be raised to indicate to the user they are trying to access something that doesn't exist.

This test should pass, currently it fails, as no error is raise when bit location 8 is attempted.

```
    def testIndex8(self):
        val = 0x12
        vector  = intbv(val)[8:]

        assert vector[7] == False
        with pytest.raises(IndexError):
            vector[8]
```

PR #399, opened 2022-12-17
Clean PyPi release flow 17/12/2022

davekeeshan:
- Change where conditional on the PyPi release is set, now at the very top of the step
- Change `make release` to `make dest` for doing the PyPi release on the github action
- Now `make release` is run on the client side machine to set the tags correctly and push to server
- Updated the setup.py to pull in the `README` file, this is used over on PyPi

To the repo owner (@josyb), once merge into the master branch in complete. Checkout from on command line and run the command:

`make release TAG=0.11.42`

or what ever rev number you want, it should :crossed_fingers: handle the rest


josyb:
Dave,

That will not work: I am not the owner, I only have write rights to the repository to merge PRs, I don't think I can push to the origin master directly
I messed up my local working copy as now my master branch is 5 commits ahead (with  dummy commits/reverts ...)

davekeeshan:
Whoops, my mistake, I just assumed you had write access to the master

I wonder if you can make another branch, call it release branch, and put tags into that?

I am not too sure how one can proceed at your level unless you have a higher privledge level, there are just some tasks that some one is required to do, and if Jan is the only person and is not available ...

davekeeshan:
Can you create a tag via the "create a new release"?

https://stackoverflow.com/questions/18216991/create-a-tag-in-a-github-repository

image

image

Inside here you can create a tag for the release, this is a way (?) of adding tags with the permissions you have? 
image

davekeeshan:
Note this is all good, i knew there would be niggles, that is why i wanted to do it manually first

PR #401, opened 2022-12-23
Raise traceSignals to DeprecationWarning from UserWarning

davekeeshan:
Missed this one when i changed analyze and verify to DeprecationWarning.  Updated code and added a test.

### DIFF
--- a/myhdl/_traceSignals.py
+++ b/myhdl/_traceSignals.py
@@ -111,7 +111,10 @@ class _TraceSignalsClass(object):
                 h = _getHierarchy(name, dut)
             else:
                 warnings.warn(
-                    "\n    traceSignals(): Deprecated usage: See http://dev.myhdl.org/meps/mep-114.html", stacklevel=2)
+                    "\n    traceSignals(): Deprecated usage: See http://dev.myhdl.org/meps/mep-114.html", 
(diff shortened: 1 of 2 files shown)


helloworld-crypto:
So how can I fix this problem: 'traceSignals(): Deprecated usage: See http://dev.myhdl.org/meps/mep-114.html'
My Python version 3.7.6 and myhdl version 0.11

josyb:
By using the correct simulation syntax, e.g.:
```python
    dft = tb_pingpongbuffer()
    dft.config_sim(trace=True, tracebackup=False, directory='vcd')
    dft.run_sim()
```

helloworld-crypto:
Thank you for promptly reply
But my testbench has mutiple return, so it returns a tuple, for example:
```
def testbench():
    ...
    return framectrl, clkgen, stimulus
```
So, my program raises another error: 'tuple' object has no attribute 'config_sim'

josyb:
Perhaps show some more code next time?

You need to decorate all code returning generators with the `@block` decorator - as you can read in [MEP114]( http://dev.myhdl.org/meps/mep-114.html)

```python
@block
def testbench():
    ...
    return framectrl, clkgen, stimulus
```

Issue #405, opened 2023-01-19
Constant value inversion (~0x1, ~0x4523) not working in VHDL


davekeeshan:
When in python some one used an inversion of a constant (~0x1, ~0x4523 etc) and `ANDs` it to another vector, the VHDL version fails
```.py
    a   = Signal(modbv(0)[8:])
    z   = Signal(modbv(0)[8:])
    
    z.next = a & ~0x1
```
This work fine in verilog as it sytax is almost identical:
```v
    assign z = (a & (~1));
```

However the current resolving for VHDL does not work:
```vhd
    z <= (a and to_unsigned(not(1), 8));
```
The not on the constan value, the 1 here, does not work

Recommend recalculating the correct inverted value and storing that instead:
```vhd
    z <= (a and to_unsigned(254, 8));
```

With special attention made to make sure that the width is correct for the assigning signal

## System information
```
MyHDL Version: master
Python Version: 3.10
```

davekeeshan:
Example of failing test [link](https://github.com/daxzio/myhdl/actions/runs/3960940762/jobs/6785777865#step:10:487)

josyb:
I rather see that the *not* is kept - to show the intent of the designer :)
You will see / have seen that the VHDL conversion is complex.

davekeeshan:
@josyb 

I respect that in some ways, but when I see it it is usually as an act of lazyness by the original coder. The whole act of generating code from code is a translation. I would veer to the side of readable code versus turgid. 

Can you send on an example of the more complex vhdl to do this?

josyb:
`z <= a and not(to_unsigned(1, 8));`
but as I said the VHDL convertor takes work to make it correct;
e.g. I am suffering to make this work: 
```python
    last_slot = Signal(False)
    next_pc = Signal(intbv(0)[WIDTHA:])

    def precalc_logic():
        lit_count = 1
        lit_offset = 0
        for i in range(OPCODES):
            if ((ir >> (i * ISA)) & ISA_MASK) == 0x10:
                lit_count = lit_count + 1
                if (i <= slot):
                    lit_offset = lit_offset + 1
        pc_offset.next = pc + lit_offset
        instr.next = (ir >> (slot * ISA)) & ISA_MASK
        last_slot.next = (slot >= (OPCODES-1))
        next_slot.next = 0 if last_slot else (slot + 1)
        next_pc.next = pc + (lit_count if last_slot else 0)
``` 
The last line converts to:
[...]

Issue #386, opened 2022-12-05
Support for switch in MyHDL


davekeeshan:
Since python 3.10, there is now support for case statements that more align with case statements in both VHDL and Verilog.

Raising this issue to track, potential, implementation in MyHDL.

[Python Switch](https://www.freecodecamp.org/news/python-switch-statement-switch-case-example/)

josyb:
Hi Dave,

I had part of this discussion with Steve, a MyHDL user in NZ already:
I copy the conversation (being a bit lazy to write a long answer)

Steve:
>Python 3.10 has a match statement - will we now get a case statement for
myHDL?? :-)

> https://learnpython.com/blog/python-match-case-statement/

Me:
>Perhaps, (probably?). And if, it will be a 'restricted' version like:
```python
match( smp ):
    case machine_states.IDLE:
        if something:
            smn.next = machine_states.BUSY
        else:
            smn.next = machine_states.IDLE 
    case machine_states.BUSY:
        ...
    case other: # optional; in VHDL we expect to cover all states explicitly; in Verilog (I don't exactly know ...)
        ... 
```
>The main (and only?) advantage is that the MyHDL code then better resembles the V* code.

Steve:
>Yes - but for decoding bit patterns also?  This is what I really want it for - creating complex if statements is less readable.  This example belwo is somewhat trivial and doesnt show the advantage clearly. but think microcontroller instruction decode for example

```python
match (concat(a,b,c)):
    case intbv(0b010):
        m.next = 0
    case intbv(0b101):
        m.next = 2
    case other:
        m.next = 3
```
Me:
>That would also work.
But for the case of microprogramming I would be tempted to try a look-up table.

Steve:
>Yes, it better resembles the v* code but it reads better than a set of if statements (IMHO), and probably compiles to a better result also.

Me:
>the 'if-elif-else chain' is converted into a V* case; the 'match', by nature, will convert to a V* case as well; but there will be no difference between the two results ... 



I have no issue with a future implementation of Python's `match` construct but I believe we can find a few other things to improve

davekeeshan:
Yes it is good to track this, 3.10 is only just starting to catch on. 

Many years ago there was a difference in RTL between synthesising a fully qualified if statement and a fully qualified case statement, but those two worlds might be a lot closer now based on what the compiler tools recognise, but having the option to do either I think is still important, for synthesis. 

https://electronics.stackexchange.com/questions/73387/difference-between-if-else-and-case-statement-in-vhdl

[...]

PR #408, opened 2023-02-03
True case statment mapping support

davekeeshan:
I have had a look at this issue, #386.  (This is an excellent issue to help one learn how the python ast works).

This works for all the conditions listed in the, basic conditional, enumerated type and concats.

This is a wholly constructive addition, it does not change any testing for code that doesn't use the python `match` keyword.  The cases where a fully populated if statement results in a `case` statement in RTL still works like that, but one should consider migrating away from that in future releases, as this is a much more important feature in RTL land (and @josyb has a preference for mapping the programmers intent with implementation [#405](https://github.com/myhdl/myhdl/issues/405#issuecomment-1397474009)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,12 @@ ANSI_GREEN=`tput setaf 2`
 ANSI_CYAN=`tput setaf 6`
 ANSI_RESET=`tput sgr0`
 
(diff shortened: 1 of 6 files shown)


josyb:
Nice!

PR #412, opened 2023-03-26
Cleaning up tests:

josyb:
removing wildcard imports
correcting indents
removing as much warnings as possible: mostly unused variables

### DIFF
--- a/myhdl/test/bugs/test_bug_1740778.py
+++ b/myhdl/test/bugs/test_bug_1740778.py
@@ -1,16 +1,12 @@
-import sys
 import os
 path = os.path
 import random
-from random import randrange
 random.seed(2)
 
-import myhdl
-from myhdl import *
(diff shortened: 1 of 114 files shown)


josyb:
Note that lot of test spit out warnings; some  *not driven* or *not read* which probably can be repaired by improving the test-benches in  the test, often there is also the *output read back* conversion warning which IMO should be removed, at least for VHDL as we use VHDL 2008 constructs in other places anyway

josyb:
@davekeeshan I decided to merge without your approval.

davekeeshan:
@josyb 

Apologies, last time I checked the tests were failing CI so I was waiting for a passing run.

josyb:
On the contrary, I owe to apologize
I had a day off and thought it would be a good thing to clear out that PR - and was a bit impatient

PR #430, opened 2024-02-22
Removed support for Python 3.7 (which is End Of Life since 2023-06-27)

josyb:
Removed stray ast.Str and the like , which throw deprecation warnings in Python 3.12

### DIFF
--- a/.github/workflows/test_checkin.yml
+++ b/.github/workflows/test_checkin.yml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false # So that one fail doesn't stop remaining tests
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.9", "3.11"]
+        python-version: ["3.8", "3.9",  "pypy-3.9", "3.10", "pypy-3.10","3.11", "3.12"]
         target: [core, iverilog, ghdl]
     runs-on: ubuntu-latest
     needs: [run_lint, build_ghdl_linux, build_iverilog_linux]
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false # So that one fail doesn't stop remaining tests
       matrix:
-        python-version: ["3.11"]
+        python-version: ["3.12"]
         #target: [core, iverilog, ghdl]
         target: [core, ghdl]
     runs-on: windows-latest
--- a/myhdl/__init__.py
+++ b/myhdl/__init__.py
@@ -49,7 +49,7 @@ traceSignals -- function that enables signal tracing in a VCD file
 toVerilog -- function that converts a design to Verilog
 
 """
-__version__ = "0.11.45"
+__version__ = "0.11.46"
 
 
 class StopSimulation(Exception):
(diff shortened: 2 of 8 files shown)


josyb:
Note that this was done to have MyHDL pass Python 3.12 warning messages regarding obsolete AST constructs

josyb:
@davekeeshan @hgomersall 
What are your ideas about this?

davekeeshan:
@josyb 

I have a weird mixed opinion about this.  In general, with python, my answer is if the rev of python has been EOLed, then it is deprecated out as soon as possible.

Personally I have moved onto 3.8 or above a long time ago, i can't live without f-strings and walrus operators.

But, and you knew there was a but coming, I was thinking that not only should 3.7 be kept, it might make sense to reintegrate 3.6 (WHAT!!!!) bare with me.  

As a FPGA/ASIC/EDA wonk that works in the corporate space, we are usually limited to whats on the system.  Most EDA tools are only rated for RH8. The problem here is that the base install of RH8 comes with Python 3.6.8.  So that means a lot of engineers, who don't know how to update python inside the corporate walls as not able to use myhdl.

Personally, I made it my mission to figure out how to compile python from source inside a big corporate, and its not pretty, but I have had the ability to roll my own for a while, but I can tell from experience that 95% of my co-workers that was outside their skill level.

However, i also made a great discovery recently [spack](https://github.com/spack/spack), which allows you to compile tools as a regular user and it compiles the dependencies as you go. It can compile any version of python that is not EOLed and a lot more, without sudo or docker access, for me it is a game changer.

The tool was designed by [Lawerence Livermore for HPC supercomputer work](https://computing.llnl.gov/projects/spack-hpc-package-manager), but I started adding in or updating some of the open EDA tools as well, so icarus, yosys, verible and verilator.

Any way that is an opinion, the simple answer is remove 3.7 support and leave a note some where telling a user the last working version for that rev.

hgomersall:
Why was 3.6 removed? Was the a specific reason? I know there are various warnings using older versions of myhdl.

josyb:
@hgomersall 
Not sure, I guess I decided so as it was EOL ( Github log : 79cf2218f4ccb998aaf012222f70d07337052d22 date:2/08/2022)

Yes, we  (royal we?) should document changes better

davekeeshan:
@josyb @hgomersall 

I think if memory serves it was actually related to the fact that github actions no longer [support 3.6](https://github.com/actions/setup-python/issues/544#issuecomment-1332535877). And that in general 3.6 is EOLed. It is just that I have spent a large portion myself fighting the old tools on the RH8 ecosystem that this is even on my radar (I know some places are still stuck on RH7, but are only now being forced to RH8 and RH7 is also going EOL)

If it has been gone a year and a half and there have been no real howls of complaint, then it is probably not an issue. With the caveat that if someone couldn't get started they might never end up here to complain.

DeflateAwning:
I'm in favor of this. If a company is stuck on Python 3.6, then maybe they should be stuck on an old version of MyHDL too.

Userspace Python installations are feasible for the rare cases where recent Python versions are inaccessible

PR #432, opened 2024-06-12
_toVerilog.py: 'skip_zero_mem_init': added *forgotten* condition

josyb:
_toVerilog.py: `'skip_zero_mem_init'`: added *forgotten* condition that the initial value must be zero to be able to skip the generation of the initial block


josyb:
Again, will merge this, notwithstanding that one test failed in PyPy3.9.
Note that it was a different test than in the previous PR

Issue #449, opened 2025-01-12
Please switch from setup-ghdl-ci to setup-ghdl


Paebbels:
> [!CAUTION]
> This GitHub Action (`setup-ghdl-ci`) has been **deprecated** in favor of the new [setup-ghdl](https://github.com/ghdl/setup-ghdl) Action.
> Please update your YAML files and transition to the new GitHub Action. See the linked repository for details on optional parameters.

**Quick example:**
```yml
jobs:
  GHDL-on-Ubuntu:
    runs-on: ubuntu-24.04
    steps:
      - name: Run VHDL Simulation
        uses: ghdl/setup-ghdl@v1
        with:
          version: nightly
          backend: mcode
```

**Why has it been changed?**  
* The Javascript GitHub Action is hard to maintain.  
  In addition, it has a high complexity with Node.js and other tools compare to a much simpler Composite Action using simple shell instructions in Bash/Powershell.
* This action doesn't support the `ubuntu-24.04` (= `ubuntu-latest`) runner image.
* This Action uses wrong tagging to select GHDL versions. See analysis: https://github.com/ghdl/ghdl/issues/2852
* Due to changes in GHDL's nightly releases and attached assets, the filenames are broken.

-----------------------
This repository has been listed by [Dependabot](https://github.com/ghdl/setup-ghdl-ci/network/dependents) as a dependent of `setup-ghdl-ci`.

davekeeshan:
@Paebbels I have made the changes to the CI flow, however the `backend: llvm`, does not appear to be recognised.  myhdl only uses the `setup-ci` from the windows configuration.

https://github.com/myhdl/myhdl/actions/runs/12756629221/job/35555176047?pr=450

![image](https://github.com/user-attachments/assets/8d31f277-0ecc-47c3-932b-5ecf9d304bad)

The documentation suggests that this is the correct way:

https://github.com/ghdl/setup-ghdl#input-parameters

Paebbels:
@davekeeshan is this your YAML code?
```
      - if: ${{ env.CI_TARGET == 'ghdl' }}
        uses: ghdl/setup-ghdl@v1
        with:
          version: nightly
          backend: llvm
```

I assume you want to run GHDL with LLVM backend within MSYS2, right? MSYS2 is setup (green arrow).  
But `setup-ghdl` sets up GHDL for Windows native, not for MSYS2 (red arrow).

[...]

PR #450, opened 2025-01-13
Update CIflow to track with changes for ghdl-ci, #449

davekeeshan:
Addressing issue raised in #449


davekeeshan:
@josyb This looks good now, i am happy with it.  I did a check though for python 3.13, there appear to be failing issues on that build, so I am not adding it with this fix.

PR #454, opened 2025-04-10
update test files for consistent seed handling

allrob23:
## Description
Updates random seed handling in test files for consistency.

### Changes
- Added `random.seed(2)` to `test_SineComputer.py`.
- Changed `random.seed = 1` to `random.seed(1)` in `test_Dec.py` and `test_LeadZeroDet.py`.

### Purpose
Ensures consistent seed usage across tests.

### DIFF
--- a/example/arith_lib/test_Dec.py
+++ b/example/arith_lib/test_Dec.py
@@ -9,7 +9,7 @@ from arith_utils import SLOW, FAST
 from Dec import Dec
 
 import random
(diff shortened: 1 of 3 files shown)


josyb:
Thanks for the PR.
I use ```random.seed('We want repeatable randomness')``` :)
There is one issue with the test-suite but that has nothing to do with your changes; we will see whether that one keeps popping up later.

PR #457, opened 2026-09-11
Vhdl enum name validation

hgomersall:
A fix to improve enum name checking, checking on keywords or invalid prefixes as well as a the enum having the same name as a port or signal.

Enum names are now stored as a `(namespace, name)` tuple which improves flexibility.


josyb:
One more thing:
At some point we have to increment the sub-minor number
Otherwise we can't publish it on PyPi

hgomersall:
I agree. Do you want that changed in this PR? How is it done?

josyb:
in myhdl/myhdl/\_\_init\_\_.py
and we should do it for every PR - even if the package, awaiting more changes, is not yet published on PyPi
I was looking on automating that - but again I am not fluent in GitHub Actions

